PR #100, opened 2026-02-22
fix(backend): raise ValueError for unknown layer types instead of ret…

### LINKED ISSUE #99: Unknown node types should throw ValueError instead of silently crashing backend
**Describe the bug**
[helper_generate_layers()](cci:1://file:///Users/ayushkumarjha/Desktop/tensormap/tensormap-backend/app/services/model_generation.py:82:0-175:27) in [model_generation.py](cci:7://file:///Users/ayushkumarjha/Desktop/tensormap/tensormap-backend/app/services/model_generation.py:0:0-0:0) is missing an `else` block for unknown layer types. It silently returns `None` for unsupported nodes, which gets appended to the model config and later causes a cryptic crash inside TensorFlow.

**To Reproduce**
1. Send an API request payload containing an unknown node type (e.g. `{"type": "invalid_layer"}`).
2. Wait for backend to process.
3. See error: TensorFlow crashes internally trying to parse the `None` config instead of throwing a clean validation error.

**Expected behavior**
The function should have an explicit `else` block that raises a `ValueError("Unknown custom layer type")`, immediately rejecting the bad payload before it reaches TensorFlow.

**Screenshots**
N/A

**Environment:**
- OS: macOS 14
- Browser: Chrome 120
- Node.js version: 18.20.8
- Python version: 3.13.9

**Additional context**

jhaayushkumar:
## Description

This PR fixes a silent failure handling issue where the `helper_generate_layers` function returned `None` when encountering an unsupported or unknown neural network layer type from the client payload. This `None` layer cascaded into the core TensorFlow parsing pipeline, causing a hard-to-trace internal server error during `tf.keras.models.model_from_json()`.

By adding an explicit `else` block to raise a `ValueError`, the backend now strictly validates layers and fails fast, giving a readable error immediately.

Fixes #99

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Ran a mock regression test locally passing an unsupported layer type (`{"type": "non_existent_layer"}`). Verified that the function correctly catches the invalid type upfront and raises a `ValueError` immediately rather than silently injecting a `None` into the model architecture.

- [ ] Existing tests pass
- [x] New tests added
- [x] Manual testing

N/A

## Checklist

- [ ] I have added/updated documentation as needed
- [x] Tests pass locally

### DIFF
--- a/tensormap-backend/app/services/model_generation.py
+++ b/tensormap-backend/app/services/model_generation.py
@@ -174,3 +174,6 @@ def helper_generate_layers(layer_params: dict) -> dict:
             "inbound_nodes": [],
         }
         return default_conv
+    
+    else:
+        raise ValueError(f"Unknown custom layer type: {layer_params.get('type')}")


jhaayushkumar:
@ivantha please review whenever convenient.

PR #109, opened 2026-02-22
fix(frontend): require only dim-1 for InputNode validation in canSaveModel (#109)

### LINKED ISSUE #108: generateModelJSON()` mutates React Flow internal state, corrupting the canvas
## Description

`generateModelJSON()` in `tensormap-frontend/src/components/DragAndDropCanvas/Helpers.jsx` performs a shallow assignment (`const filteredJSON = model_data`) and then mutates the object in-place by deleting `node.width`, `node.position`, `node.positionAbsolute`, `edge.id`, etc. Because JavaScript objects are passed by reference, this **directly mutates React Flow's internal store**, causing the canvas to break after model validation — nodes lose their positions and edges lose their identifying IDs.

### Related bugs in the same file

1. **Wrong validation param keys in `enableValidateButton()`:**
   - `custominput` validation checks `dim-x`, `dim-y`, and `units` — but InputNode actually uses `dim-1`, `dim-2`, `dim-3` (see `InputNode.jsx`).
   - `customflatten` validation checks for `units` — but FlattenNode has no `units` property (only `dim-x`, `dim-y`).
   - This means the validate button never correctly gates on InputNode or FlattenNode parameters.

2. **Canvas.jsx `.catch()` block replaces entire `formState`:**
   - `setFormState({ modelValidatedSuccessfully: false, modalContent: error.message })` throws away all form field values instead of merging with spread.

## Steps to Reproduce

1. Open the Deep Learning canvas
2. Drag an InputNode and a DenseNode onto the canvas
3. Connect them, fill in all form fields
4. Click Validate
5. **Observe:** After validation, the canvas nodes lose their positions; the graph is visually corrupted

## Expected Behavior

- The canvas should remain intact after validation
- Node positions, dimensions, and edge IDs should be preserved
- Validate button should be disabled when InputNode `dim-1`, `dim-2`, `dim-3` are empty

## Affected Files

- `tensormap-frontend/src/components/DragAndDropCanvas/Helpers.jsx`
- `tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx`

## Severity

**Critical** - the core model-building workflow is broken; users cannot validate models without corrupting the canvas.

kartikeyg0104:
## Summary

`canSaveModel` in `Helpers.jsx` was incorrectly requiring all three InputNode dimensions (`dim-1`, `dim-2`, `dim-3`) before enabling the save/validate button. This blocked users with 1D or 2D input shapes from ever saving their model.

`InputNode.jsx` already filters out empty dimensions before display, making `dim-2` and `dim-3` explicitly optional. Only `dim-1` is required.

## Root Cause

After rebasing onto the updated `main` branch (which already included the deep-clone fix for `generateModelJSON` and the `Canvas.jsx` `.catch()` spread fix), the only remaining change in this PR is the InputNode dimension validation. The previous iteration of this PR required all three dimensions, which contradicts the optional nature of `dim-2` and `dim-3`.

### `canSaveModel` — InputNode validation

**Before:**
```js
} else if (node.type === "custominput") {
  // Validate all three dimensions
  if (
    !node.data.params["dim-1"] || node.data.params["dim-1"] === "" ||
    !node.data.params["dim-2"] || node.data.params["dim-2"] === "" ||
    !node.data.params["dim-3"] || node.data.params["dim-3"] === ""
  ) {
    return false;
  }
}
```

**After:**
```js
} else if (node.type === "custominput") {
  // Only dim-1 is required; dim-2 and dim-3 are optional (e.g. 1D / 2D inputs)
  if (!node.data.params["dim-1"] || node.data.params["dim-1"] === "") {
    return false;
  }
}
```

This matches `InputNode.jsx`, which renders only the non-empty dimensions:
```js
const dims = [data.params["dim-1"], data.params["dim-2"], data.params["dim-3"]]
  .filter((d) => d !== "" && d !== undefined)
  .join(" × ");
```

## Testing

- Drop an InputNode, fill only `dim-1` → validate button should become active
- Drop an InputNode, leave `dim-1` empty → validate button should remain disabled
- Drop an InputNode, fill all three dims → validate button should become active

## Related

Closes #108

### DIFF
--- a/tensormap-frontend/src/components/DragAndDropCanvas/Helpers.jsx
+++ b/tensormap-frontend/src/components/DragAndDropCanvas/Helpers.jsx
@@ -17,6 +17,7 @@ export const canSaveModel = (modelName, modelData) => {
         return false;
       }
     } else if (node.type === "custominput") {
+      // Only dim-1 is required; dim-2 and dim-3 are optional (e.g. 1D / 2D inputs)
       if (!node.data.params["dim-1"] || node.data.params["dim-1"] === "") {
         return false;
       }


ivantha (changes requested):
Please rebase. The branch is stable now. Sorry about the inconvenience.

ivantha (commented):
1. PR description is overstated : The description claims 3 fixes (deep-clone in generateModelJSON, FlattenNode validation, and Canvas.jsx .catch() fix), but the diff only contains 1: adding dim-2/dim-3 validation for custominput. The other two are already fixed on main. Please update the description to match the actual change.
2. Are all 3 dimensions truly required? : InputNode.jsx filters out empty dimensions for display, suggesting they're optional (e.g., a 1D input only needs dim-1). This validation now requires all three, which would block users from saving 1D or 2D input shapes. Is that intentional?

ivantha:
Hey, thanks for the PR! However, the code change here actually does the opposite of what the description says. The PR says it should "require only dim-1," but the diff adds dim-2 and dim-3 as required checks too. This would break things for users with 1D or 2D inputs since the save button would never enable. The main branch already has the correct validation (only dim-1 required), so this change would be a step backwards. Closing this one out.

kartikeyg0104:
The bug doesn't exist. The validation is already correct on main in the recent prs.

PR #111, opened 2026-02-22
perf(backend): cache CSV column names in DB instead of reading from disk on every list call

### LINKED ISSUE #110: `get_all_files_service` reads every CSV from disk on every list call — O(n × file_size) I/O per request
## Description

In `tensormap-backend/app/services/data_upload.py`, the `get_all_files_service()` function iterates over **every** `DataFile` record and calls `pd.read_csv()` on each CSV file just to extract column names. This creates severe performance and reliability problems:

1. **O(n × file_size) disk I/O per request:** Every call to the "list files" endpoint reads every uploaded CSV from disk. With 10 files of 50MB each, that's 500MB of disk reads per API call.

2. **Event loop blocking:** The `async def` handler calls this synchronous pandas I/O, blocking the entire uvicorn event loop during the operation.

3. **Fragile:** If any file is deleted from disk (but the DB record still exists), or if a CSV is malformed, the entire endpoint returns a 500 error - one bad file breaks listing for all files.

4. **No caching:** Column names never change after upload, yet they're re-read from disk every time.

## Expected Behavior

Column names should be extracted **once at upload time** and stored in the database. The list endpoint should be a single DB query with zero disk I/O.

## Proposed Solution

1. Add a nullable `columns` JSON field to the `DataFile` model
2. Populate it at upload time by calling `pd.read_csv()` once
3. `get_all_files_service` reads `columns` from the DB record
4. For existing files without cached columns, do a one-time backfill read and update the record

## Affected Files

- `tensormap-backend/app/models/data.py`
- `tensormap-backend/app/services/data_upload.py`
- New Alembic migration

## Severity

**High** - degrades server performance linearly with the number of uploaded files; causes full server stall under load.

kartikeyg0104:
## Problem(Fixes #110)

`get_all_files_service()` iterates over every `DataFile` record and calls `pd.read_csv()` on each CSV to extract column names. This causes:

- **O(n × file_size) disk I/O per request** — every list-files API call reads every CSV from disk
- **Event loop blocking** — synchronous pandas I/O inside async handlers
- **Fragility** — one missing/corrupt file on disk breaks listing for all files (returns 500)
- **No caching** — column names never change after upload, yet they're re-read every time

### `app/models/data.py`
- Add nullable `columns: List[str] | None` JSON field to `DataFile` model

### `app/services/data_upload.py`
- **`add_file_service()`**: Extract column names from CSV at upload time and store in the `columns` field
- **`get_all_files_service()`**: Read `columns` from the DB record instead of `pd.read_csv()` on every call. Falls back to disk read only for existing files without cached columns, and backfills the DB record on first access.

### Migration
- `f1a2b3c4d5e6_add_columns_json_to_data_file.py`: Adds `columns` JSON column to `data_file` table (nullable, backward-compatible)

## Testing

1. Upload a CSV → verify `columns` is populated in the DB
2. `GET /data/upload/file` → should return column names without reading from disk
3. Existing files without cached columns → should be backfilled on first list call
4. Upload a ZIP → `columns` should be `null` (no CSV to parse)


ivantha (changes requested):
1. row_count is broken for cached files. When columns come from the DB, df is None, so row_count is always 0. The
  frontend still shows this value in the UI (Rows: {dataset.rows.toLocaleString()}). You'd need to either cache the
  row count in a new DB field too, or intentionally drop it from the response/UI.
  2. Ruff is failing. List[str] should be list[str] (project targets 3.12+), and the imports are unsorted in both
  changed files. Running ruff check --fix . && ruff format . should take care of it.
  3. Small inconsistency: the upload path uses nrows=0 to read just the header, but the backfill path in
  get_all_files_service reads the entire CSV. Probably worth making the backfill use nrows=0 too, since the row
  count is lost after the first access anyway

ivantha (changes requested):
Non-CSV files hit the backfill path on every request. Since the upload path only populates columns for CSVs, ZIP files will always have columns=None. That means the backfill branch runs every time, tries pd.read_csv() on a ZIP, fails, and never persists the result. So you get repeated failed disk reads on every list call for every ZIP file, which is actually worse than before.

Fix: wrap the backfill in if file.file_type == "csv":, and for non-CSV files just default to fields=[], row_count=0 without touching disk.

kartikeyg0104:
> Non-CSV files hit the backfill path on every request. Since the upload path only populates columns for CSVs, ZIP files will always have columns=None. That means the backfill branch runs every time, tries pd.read_csv() on a ZIP, fails, and never persists the result. So you get repeated failed disk reads on every list call for every ZIP file, which is actually worse than before.
> 
> Fix: wrap the backfill in if file.file_type == "csv":, and for non-CSV files just default to fields=[], row_count=0 without touching disk.

Screenshot 2026-02-28 at 4 23 08 PM

Good catch, thanks @ivantha.
You are right I was only checking if file.columns was not None before triggering the backfill. I didn’t think about the fact that non-CSV files (like ZIPs) wouldn’t have columns at upload time so they did end up hitting pd.read_csv() on every list call and just failing quietly. That’s definitely worse than before my bad.

Fixed in 408e864. The backfill now only runs when file.file_type == "csv". Non-CSV files default to fields=[] and row_count=0 without touching disk. Ruff is clean.

ivantha (approved):
Lgtm 👍🏼 Please squash your commits and rebase.

PR #155, opened 2026-02-25
feat: Implement dataset pagination in backend API and frontend display

### LINKED ISSUE #136: Paginate Dataset Preview Table
**Difficulty:** Medium
**Area:** full stack

## Description

The dataset preview functionality in TensorMap loads and displays all rows of a CSV file at once. For small datasets this works fine, but for larger files (thousands or tens of thousands of rows), it causes performance issues: slow rendering, high memory usage in the browser, and a poor user experience with an overwhelmingly long table.

This issue adds server-side pagination to the dataset preview. The backend will accept `page` and `page_size` query parameters and return only the requested slice of data along with pagination metadata. The frontend will display 50 rows per page with previous/next navigation buttons.

## Acceptance Criteria

- [ ] Backend endpoint accepts `page` and `page_size` query parameters
- [ ] Response includes pagination metadata: `page`, `page_size`, `total_rows`, `total_pages`
- [ ] Frontend displays 50 rows per page by default
- [ ] Previous/Next buttons navigate between pages
- [ ] Previous button is disabled on the first page; Next button is disabled on the last page
- [ ] Page indicator shows "Page X of Y" and "Showing N to M of Total rows"
- [ ] Initial page load is fast even for large datasets (thousands of rows)
- [ ] Empty datasets display an appropriate message instead of pagination controls

## Resources

- `tensormap-frontend/src/components/Process/DisplayDataset.jsx` — dataset preview component (or equivalent)
- `tensormap-backend/app/routers/data_process.py` — data processing endpoints
- `tensormap-backend/app/routers/data_upload.py` — data upload/listing endpoints
- `tensormap-backend/app/models/data.py` — `DataFile` model
- `tensormap-frontend/src/services/FileServices.js` — file API service calls
- `tensormap-frontend/src/components/ui/button.jsx` — shadcn Button component

anudeeps0306:
## Description

This PR introduces backend and frontend pagination support for dataset previews, significantly improving performance, payload size, and user experience when working with large files.

###  Backend Updates

* **Endpoint Update (`data_process.py`):**
  Enhanced `GET /data/process/file/{file_id}` to accept `page` (default: 1) and `page_size` (default: 50) as query parameters.
* **Service Layer Optimization (`data_process.py`):**
  The raw DataFrame is now sliced using `.iloc[start_idx:end_idx]` **before** converting to JSON, reducing marshalling time and response size.
* **Response Structure Modification:**
  Replaced the simple string payload with a structured response containing:

  * `data` array
  * Pagination metadata: `page`, `page_size`, `total_rows`, `total_pages`

### Frontend Enhancements

* **Service Enhancement (`FileServices.js`):**
  Now passes `page` and `page_size` in API requests and returns the full pagination-aware response object.
* **Preview Component Rewrite (`DisplayDataset.jsx`):**

  * Uses shadcn UI `Button` components for **Previous** and **Next** controls
  * Dynamically renders pagination info: **"Page X of Y"** and **"Showing N to M of Total rows"**
  * Handles loading states and shows a graceful empty state when no rows are returned
  * Automatically manages boundaries (disables **Previous** on first page and **Next** on last page)

### Verification Details

* **Application Build Integrity:**
  Ran `npm run build` (Vite) successfully to confirm all updated JSX and modules compile without syntax or runtime errors.
* **Pagination Defaults:**
  Verified the API slices and returns **50 rows by default**, matching the acceptance criteria.

Fixes #136

---

## Type of Change

* [ ] Bug fix
* [x] New feature
* [ ] Breaking change
* [ ] Documentation update

---

## How Has This Been Tested?

* Verified backend API returns correct paginated slices and metadata

* Tested frontend pagination controls (Previous/Next) and boundary conditions

* Verified empty state handling when no rows are returned

* Confirmed successful production build with Vite (`npm run build`)

* [x] Existing tests pass

* [ ] New tests added

* [x] Manual testing

---

image
image

---

## Checklist

* [x] My code follows the project's style guidelines
* [x] I have performed a self-review
* [ ] I have added/updated documentation as needed
* [x] My changes generate no new warnings
* [x] Tests pass locally

---


ivantha (approved):
Lgtm 👍🏼 Please rebase.

ivantha:
## ◎ Argus Review

This PR successfully implements server-side pagination for dataset previews, and the core approach is sound — slicing before JSON serialization, returning structured pagination metadata, and updating the frontend to consume it. However, there are two notable bugs: the frontend does not reset the page when `fileId` changes (causing wrong-page fetches on dataset switching), and concurrent rapid page changes can result in a race condition showing stale data. The backend performance opportunity is partially missed because the full CSV is still read into memory before slicing. Additionally, the delete-file behavior change is an undocumented breaking fix that silently diverges from prior semantics.

### Findings (15)

#### 🟠 Page state not reset when fileId changes
**Severity:** high | **Category:** bug | **File:** `tensormap-frontend/src/components/Process/DisplayDataset.jsx:41`

When `fileId` changes (e.g., the user selects a different dataset), the `page` state is **not** reset to `1`. The `useEffect` depends on both `[fileId, page, fetchData]`, so when `fileId` changes, it fires `fetchData(fileId, page)` with whatever page the user was on for the *previous* dataset. If the new dataset is shorter, this results in an out-of-range page request, and the backend returns an empty data list — the component then shows 'Dataset is empty' instead of the first page.

```jsx
useEffect(() => {
  if (fileId) {
    fetchData(fileId, page); // ← page may be 3, 4, etc. from prior dataset
  }
}, [fileId, page, fetchData]);
```

**Suggestion:**
```
Add a separate `useEffect` that resets `page` to `1` when `fileId` changes:
```jsx
useEffect(() => {
  setPage(1);
}, [fileId]);

useEffect(() => {
  if (fileId) {
    fetchData(fileId, page);
  }
}, [fileId, page, fetchData]);
```
Alternatively, reset page inside the first `useEffect` before calling `fetchData` and use a ref to track `fileId` changes.
```

---

#### 🟠 Race condition: rapid page changes can render stale data
**Severity:** high | **Category:** bug | **File:** `tensormap-frontend/src/components/Process/DisplayDataset.jsx:20`

If the user clicks **Next** or **Previous** quickly, multiple concurrent `getFileData` requests are in flight simultaneously. React state updates are applied in the order responses arrive — not the order requests were sent. A slower earlier request can overwrite the result of a faster later request, causing the displayed data to show the wrong page.

For example:
1. User navigates: page 1 → 2 → 3
2. Requests for pages 1, 2, 3 are all in flight
3. Page 2 responds last → UI shows page 2 data but `pagination.page` says 3

**Suggestion:**
```
Use `AbortController` to cancel in-flight requests when a new one starts:
```jsx
const fetchData = useCallback(async (fileId, currentPage = 1) => {
  if (abortControllerRef.current) {
    abortControllerRef.current.abort();
  }
  const controller = new AbortController();
  abortControllerRef.current = controller;
  setIsLoading(true);
  try {
    setError(null);
    const fileData = await getFileData(fileId, currentPage, 50, controller.signal);
    if (!controller.signal.aborted) {
      setData(fileData.data);
      setPagination(fileData.pagination);
    }
  } catch (e) {
    if (e.name !== 'AbortError') {
      logger.error('Error fetching dataset:', e);
      setError('Failed to load dataset');
      setData(null);
    }
  } finally {
    if (!controller.signal.aborted) setIsLoading(false);
  }
}, []);
```
```

---

#### 🟠 Entire CSV file loaded into memory for every paginated request
**Severity:** high | **Category:** performance | **File:** `tensormap-backend/app/services/data_process.py:209`

Despite the PR's goal of reducing payload size and improving performance, the full DataFrame is still read into memory via `pd.read_csv(path)` before slicing with `.iloc[start_idx:end_idx]`. For a 500 MB CSV with 5 million rows, every page request — even for page 1 of 50 rows — allocates and parses the entire file.

This means the backend memory and CPU overhead per request is **identical** to the original non-paginated implementation. Only the response size (network transfer) is reduced.

**Suggestion:**
```
Use `pd.read_csv`'s built-in `skiprows` and `nrows` parameters to read only the target slice:
```python
# Compute skip/nrows before reading
skip = list(range(1, start_idx + 1))  # skip rows after header
df_page = pd.read_csv(path, skiprows=skip if start_idx > 0 else None, nrows=page_size)
```
For `total_rows`, maintain a separate lightweight row-count (e.g., cached metadata in the DB, or count lines with `sum(1 for _ in open(path)) - 1`). Alternatively, use `dask` for lazy evaluation on large files.
```

---

#### 🟡 Out-of-range page number silently returns empty data with HTTP 200
**Severity:** medium | **Category:** bug | **File:** `tensormap-backend/app/services/data_process.py:228`

If a client requests a page beyond the last available page (e.g., `page=999` for a 50-row file), `start_idx` exceeds `total_rows`, `df.iloc[start_idx:end_idx]` returns an empty DataFrame, and the response is HTTP `200` with `data: []`. There is no client-facing indication that the requested page is invalid — the response is indistinguishable from a genuinely empty dataset.

This makes it hard to detect pagination bugs (like the `fileId`-change bug above) because the client sees `200 OK` with empty data.

**Suggestion:**
```
Add an explicit guard after computing `total_pages`:
```python
if total_rows > 0 and page > total_pages:
    return _resp(400, False, f"Page {page} exceeds total pages ({total_pages})")
```
Alternatively, clamp the page to the valid range and include a `requested_page` field in the pagination metadata so callers can detect clamping.
```

---

#### 🟡 Silent behavior change: DB record deleted even when file is missing from disk
**Severity:** medium | **Category:** logic | **File:** `tensormap-backend/app/services/data_upload.py:131`

The old logic returned `400 'File not found'` and left the DB record intact if the file did not exist on disk. The new logic always deletes the DB record and commits — regardless of whether the file existed on disk:

```python
# Before:
if os.path.isfile(file_path):
    os.remove(file_path)
    db.delete(file)
    db.commit()
    return _resp(200, True, "File deleted successfully")
else:
    return _resp(400, False, "File not found")   # ← DB untouched

# After:
if os.path.isfile(file_path):
    os.remove(file_path)

db.delete(file)   # ← always runs
db.commit()
return _resp(200, True, "File deleted successfully")
```

This is a **breaking behavioral change** that is not documented in the PR description. Callers who previously received `400` to detect inconsistent state now silently get `200`, masking orphaned-record cleanup as normal deletion.

**Suggestion:**
```
If the intent is to clean up orphaned DB records (valid use case), document this explicitly and choose one of:
1. Return a distinct `207 Multi-Status` or a different message like `"File metadata deleted (disk file was already missing)"`.
2. Keep the old `400` guard but add a separate endpoint for force-deleting orphaned records.
3. At minimum, log a warning: `logger.warning('File not found on disk for id %s, deleting DB record anyway', file_id)`.
```

---

#### 🟡 Unnecessary double JSON serialization/deserialization round-trip
**Severity:** medium | **Category:** performance | **File:** `tensormap-backend/app/services/data_process.py:235`

The code converts the DataFrame to a JSON string and then immediately parses it back to a Python list:
```python
data_json = df_page.to_json(orient="records")  # DataFrame → JSON string
data_list = json.loads(data_json)               # JSON string → Python list
```
This double pass wastes CPU cycles and memory allocating an intermediate string. For a 50-row page this is trivial, but it reflects an unnecessary pattern that would become more costly at `page_size=1000`.

**Suggestion:**
```
Use `DataFrame.to_dict()` directly to get the Python list without an intermediate JSON string:
```python
data_list = df_page.to_dict(orient="records")
```
Note: `to_dict` preserves native Python types (int, float, etc.) rather than JSON-serialized strings, which is more correct for downstream use.
```

---

#### 🟡 Dead code: string type guard for fileData.data is unreachable
**Severity:** medium | **Category:** bug | **File:** `tensormap-frontend/src/components/Process/DisplayDataset.jsx:26`

The component contains a defensive check that will never be true given the current backend implementation:
```jsx
let parsedData;
if (typeof fileData.data === "string") {
  parsedData = JSON.parse(fileData.data);  // ← dead branch
} else {
  parsedData = fileData.data;              // ← always taken
}
```
The backend always returns `data` as a parsed JSON array (via `json.loads(data_json)`), not a string. `FileServices.js` returns `response.data` (Axios already parses JSON), so `fileData.data` will always be a JavaScript array. This dead branch adds confusion for future maintainers.

**Suggestion:**
```
Remove the unnecessary branch:
```jsx
const parsedData = fileData.data;
setData(parsedData);
setPagination(fileData.pagination);
```
If you need defensive coding for an unexpected API shape, add a runtime type assertion or use a validation library (e.g., Zod) rather than silent fallback parsing.
```

---

#### 🟡 Success path bypasses the _resp() helper, breaking response structure consistency
**Severity:** medium | **Category:** maintainability | **File:** `tensormap-backend/app/services/data_process.py:238`

All other code paths in `get_file_data` (and throughout the service layer) use the `_resp()` helper to produce a consistent `(dict, status_code)` tuple. The new success path constructs the response dict manually:
```python
body = {
    "success": True,
    "message": "Data sent successfully",
    "data": data_list,
    "pagination": {...},
}
return body, 200
```
This diverges from the established pattern, making the codebase inconsistent and harder to maintain. If `_resp()` is ever updated (e.g., to add a `timestamp` field), the paginated response will be excluded.

**Suggestion:**
```
Either extend `_resp()` to support extra keys, or create a new helper:
```python
def _paginated_resp(data: list, pagination: dict) -> tuple:
    body = {
        "success": True,
        "message": "Data sent successfully",
        "data": data,
        "pagination": pagination,
    }
    return body, 200

# Then:
return _paginated_resp(data_list, {"page": page, "page_size": page_size, "total_rows": total_rows, "total_pages": total_pages})
```
```

---

#### 🟡 Route changed to async def but service calls may be synchronous/blocking
**Severity:** medium | **Category:** bug | **File:** `tensormap-backend/app/routers/data_upload.py:36`

The upload route was changed from `def` to `async def`:
```python
# Before:
def upload_file(data: UploadFile, ...)
# After:
async def upload_file(data: UploadFile = File(...), ...)
```
FastAPI runs synchronous (`def`) route handlers in a thread pool executor to avoid blocking the event loop. By changing to `async def`, all code in the handler now runs directly on the event loop. If the underlying service functions perform blocking I/O (disk writes, `shutil.copyfileobj`, `os.rename`, etc.), they will block the event loop and degrade performance under concurrent load.

**Suggestion:**
```
Only use `async def` if the handler actually uses `await` for I/O operations. If the service functions are synchronous:
- Revert to `def upload_file(...)` to restore thread pool execution, OR
- Convert the service layer to be fully async using `aiofiles` and `await file.read()`.

Also note: `File(...)` is redundant for `UploadFile` parameters — FastAPI infers it automatically. The original `data: UploadFile` (without `= File(...)`) is equivalent and cleaner.
```

---

#### 🟡 No new tests for pagination logic
**Severity:** medium | **Category:** testing | **File:** `tensormap-backend/app/services/data_process.py:209`

The PR adds non-trivial pagination logic with multiple edge cases that are untested:
- Page 1 (first page)
- Last page (partial slice)
- `total_rows` exactly divisible by `page_size`
- `total_rows = 0` (empty file)
- `page > total_pages` (out-of-range — currently returns 200 silently)
- `page_size = 1` (single row per page)
- `page_size = total_rows` (one-page result)

The PR checklist marks 'New tests added' as unchecked, which is a concern given the number of edge cases.

**Suggestion:**
```
Add unit tests for `get_file_data` in the test suite, covering at minimum:
```python
def test_get_file_data_page1(mock_db, tmp_csv):
    body, status = get_file_data(mock_db, file_id=..., page=1, page_size=2)
    assert body["pagination"]["page"] == 1
    assert len(body["data"]) == 2

def test_get_file_data_last_page_partial(mock_db, tmp_csv):
    # CSV has 5 rows, page_size=2, page=3 → 1 row
    body, status = get_file_data(mock_db, file_id=..., page=3, page_size=2)
    assert len(body["data"]) == 1

def test_get_file_data_out_of_range_page(mock_db, tmp_csv):
    body, status = get_file_data(mock_db, file_id=..., page=999, page_size=50)
    # Should return error, not 200 with empty data
    assert status == 400
```
```

---

#### 🔵 Docstrings not updated to reflect pagination behavior
**Severity:** low | **Category:** documentation | **File:** `tensormap-backend/app/routers/data_process.py:83`

Two docstrings still describe the old behavior:
1. `routers/data_process.py`: `"""Return the full contents of a CSV file as JSON records."""` — the endpoint no longer returns the full contents.
2. `services/data_process.py`: `"""Read and return the full contents of a CSV file as JSON."""` — same issue.

This misleads future developers about what the function/endpoint does.

**Suggestion:**
```
Update both docstrings:
```python
# router:
"""Return a paginated slice of a CSV file's contents with pagination metadata."""

# service:
"""Read and return a paginated slice of a CSV file as structured JSON.

Args:
    page: 1-based page number.
    page_size: Number of rows per page (1–1000).

Returns:
    Tuple of (response_body_dict, status_code) where body includes
    'data' (list of row dicts) and 'pagination' metadata.
"""
```
```

---

#### 🔵 Page size hardcoded to 50 in component without configuration
**Severity:** low | **Category:** maintainability | **File:** `tensormap-frontend/src/components/Process/DisplayDataset.jsx:22`

The page size `50` appears as a magic number in `fetchData`:
```jsx
const fileData = await getFileData(fileId, currentPage, 50);
```
This is also the default in `FileServices.js` and the backend. If the page size ever needs to change (e.g., for performance tuning or UX reasons), it must be updated in multiple places.

**Suggestion:**
```
Define a named constant at the top of the file or accept it as a prop:
```jsx
const PAGE_SIZE = 50;

// Or as prop:
const DisplayDataset = ({ fileId, pageSize = 50 }) => {
  ...
  const fileData = await getFileData(fileId, currentPage, pageSize);
```
If the component accepts `pageSize` as a prop, callers can customize the preview density without touching the component internals.
```

---

#### 🔵 Array index used as React key for table rows
**Severity:** low | **Category:** maintainability | **File:** `tensormap-frontend/src/components/Process/DisplayDataset.jsx:82`

Both `<tr>` and `<td>` elements use array index as key:
```jsx
{data.map((row, index) => (
  <tr key={index} className="border-b">
    {Object.values(row).map((value, idx) => (
      <td key={idx} ...>
```
Using array indices as keys is generally discouraged in React. While it's often unavoidable for CSV data without unique identifiers, it can cause subtle rendering bugs when data is sorted, filtered, or updated — React may re-use DOM nodes incorrectly, leading to mismatched cell values.

**Suggestion:**
```
If the CSV rows don't have a natural unique key, consider a composite key using the page and index:
```jsx
<tr key={`row-${pagination?.page}-${index}`}>
  {Object.values(row).map((value, idx) => (
    <td key={`cell-${pagination?.page}-${index}-${idx}`}>
```
This ensures keys are unique across page changes, preventing incorrect DOM reuse during page transitions.
```

---

#### ℹ️ Positive: Null value handling in table cells is improved
**Severity:** info | **Category:** style | **File:** `tensormap-frontend/src/components/Process/DisplayDataset.jsx:75`

The original code rendered `{value}` directly, which would silently render `null` as blank but could cause React warnings for `undefined` and could render boolean `false` incorrectly. The new code:
```jsx
{value !== null ? String(value) : ""}
```
explicitly converts values to strings, handling `null` with an empty string. This is a clear improvement in robustness.

**Suggestion:**
```
Consider also handling `undefined` for robustness, since CSV parsers may introduce `undefined` for missing columns:
```jsx
{value != null ? String(value) : ""}
// Uses loose != to catch both null and undefined
```
```

---

#### ℹ️ Positive: Proper loading state management with isLoading flag
**Severity:** info | **Category:** style | **File:** `tensormap-frontend/src/components/Process/DisplayDataset.jsx:15`

The original component rendered the 'Loading data...' message only in the `!data || data.length === 0` branch, which meant a loaded-but-empty dataset would show the loading message. The new implementation adds a dedicated `isLoading` boolean state, correctly separating the loading, empty, error, and data-available states. The `finally` block in `fetchData` ensures `isLoading` is always set to `false` even on error. This is a solid UX improvement.

**Suggestion:**
```
No changes needed — this is well-implemented. As an optional enhancement, you could also disable the pagination buttons while `isLoading` is true to prevent double-clicks during a page transition:
```jsx
<Button
  disabled={isLoading || !pagination || pagination.page <= 1}
  ...
>
  Previous
</Button>
```
```

---


*◎ Generated by [Argus](https://github.com) — AI-Powered PR Review*

anudeeps0306:
@ivantha, made the changes

ivantha:
@anudeeps0306 Please squash all the commits into one.

anudeeps0306:
@ivantha @rajutkarsh07 i have created new pull request #241 with same changes.

PR #159, opened 2026-02-25
feat(fullstack): add grouped loss functions with auto-select based on…

### LINKED ISSUE #142: [Feature] Let users choose the loss function on the Training page
**Difficulty:** Medium
**Area:** full stack

## Description
The loss function is never a user choice. `create_model_service` and `set_model_config_service` in `tensormap-backend/app/services/deep_learning.py` both derive it from `ProblemType` (lines 186-189 and 380-383), storing either `sparse_categorical_crossentropy` or `mean_squared_error` on `ModelBasic.loss`, and `model_run.py:308-311` branches on that single string, silently falling back to `MeanSquaredError()` for anything it does not recognise. The Training page has no loss control at all: `tensormap-frontend/src/containers/Training/Training.jsx` only renders the stored value read-only in the history table at :963. Binary classification, multi-label classification and outlier-robust regression are therefore unreachable from the UI.

## Acceptance Criteria
- [ ] The Training page gains a Loss Function `<Select>` alongside the existing Optimizer and Result Metrics selects (Training.jsx:741-787), backed by an option list in the same shape as `optimizerOptions` (:44-56)
- [ ] Options are grouped Classification and Regression, covering at least `SparseCategoricalCrossentropy`, `CategoricalCrossentropy`, `BinaryCrossentropy`, `MeanSquaredError`, `MeanAbsoluteError` and `Huber`
- [ ] Choosing a problem type preselects the matching default loss, and the user can override that default
- [ ] `TrainingConfigRequest` (`tensormap-backend/app/schemas/deep_learning.py:97-109`) gains a validated `loss` field and rejects unknown names with a clear 4xx message
- [ ] `set_model_config_service` persists the submitted loss instead of recomputing it from `ProblemType`, while `create_model_service` keeps a problem-type default so models created before the field existed still train
- [ ] `model_run.py` resolves the stored name through one name-to-Keras-loss mapping rather than the two-branch `if`, and raises on an unrecognised name instead of quietly compiling with `MeanSquaredError`
- [ ] Backend tests cover the mapping for every supported name plus rejection of an unknown one

## Resources
- `tensormap-backend/app/services/deep_learning.py:186-189` - loss derived from `ProblemType` when the model is created
- `tensormap-backend/app/services/deep_learning.py:380-383` - the same derivation repeated when the training config is saved
- `tensormap-backend/app/services/model_run.py:308-316` - the two-branch compile step, the only place the stored string is read
- `tensormap-backend/app/models/ml.py:33` - `ModelBasic.loss`, already a nullable 50-char column, so the persisted value itself needs no migration
- `tensormap-backend/app/schemas/deep_learning.py:97-109` - `TrainingConfigRequest`, currently validating `optimizer` and `metric` but no loss
- `tensormap-backend/app/shared/enums.py:4-9` - `ProblemType`, the enum the default should key off
- `tensormap-frontend/src/containers/Training/Training.jsx:44-56` - `optimizerOptions` and `metricOptions`, the pattern a `lossOptions` list should follow
- `tensormap-frontend/src/containers/Training/Training.jsx:741-787` - the Optimizer and Result Metrics selects the new control sits beside
- [Keras loss functions](https://www.tensorflow.org/api_docs/python/tf/keras/losses)

NamanmeetSingh:
## Description

This PR introduces comprehensive loss function configurations to both the React frontend and FastAPI backend, replacing the hardcoded crossentropy/mse logic.

**Implementation Details:**
* **Frontend:** Added a grouped `<Select>` component for Loss Functions to the `Training.jsx` UI. Bound an `onValueChange` listener to the Problem Type dropdown so that selecting Classification defaults the loss to `sparse_categorical_crossentropy`, and Regression defaults to `mean_squared_error`.
* **Backend Schema:** Introduced a strict `LossFunction = Literal[...]` type hint in `deep_learning.py` for both `DLModelConfig` and `TrainingConfigRequest`. This offloads invalid name rejection to Pydantic, which generates clear 422 HTTP validation errors automatically.
* **Backend Compilation:** Refactored the compile step in `model_run.py` to use a mapping dictionary referencing the 6 specific `tf.keras.losses` objects.

Fixes #142 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [x] **Frontend State Test:** Verified that toggling between Classification and Regression auto-populates the Loss Function dropdown correctly, and manual overrides are successfully stored in the React state.
- [x] **API Payload Test:** Monitored the network tab to ensure the `loss` string is successfully injected into the `updateTrainingConfig` request payload.
- [x] **Backend Validation Test:** Triggered the training endpoint via Swagger Docs (`/docs`) with an invalid loss name to verify Pydantic correctly rejects it with a clear `422` error.
- [x] **Training Test:** Verified that TensorFlow successfully compiles and trains the model utilizing the dynamically mapped Keras loss object.

## Checklist

- [x] Tests pass locally

### DIFF
--- a/tensormap-backend/app/schemas/deep_learning.py
+++ b/tensormap-backend/app/schemas/deep_learning.py
@@ -1,9 +1,19 @@
 """Request schemas for model validation, code generation, and training endpoints."""
 
 import uuid as uuid_pkg
-from typing import Any
+from typing import Any, Literal
 
 from pydantic import BaseModel, Field
+
+# Define our strict allowed loss functions
+LossFunction = Literal[
+    "sparse_categorical_crossentropy",
+    "categorical_crossentropy",
+    "binary_crossentropy",
+    "mean_squared_error",
+    "mean_absolute_error",
+    "huber",
+]
 
 
 # --- "code" sub-models ---
@@ -22,6 +32,7 @@ class DLModelConfig(BaseModel):
     optimizer: str = Field(min_length=1)
     metric: str = Field(min_length=1)
     epochs: int = Field(gt=0)
+    loss: LossFunction = "sparse_categorical_crossentropy"
 
 
 class CodeConfig(BaseModel):
@@ -107,3 +118,4 @@ class TrainingConfigRequest(BaseModel):
     epochs: int = Field(gt=0)
     batch_size: int = Field(default=32, gt=0)
     project_id: uuid_pkg.UUID | None = None
+    loss: LossFunction = "sparse_categorical_crossentropy"
--- a/tensormap-frontend/src/containers/Training/Training.jsx
+++ b/tensormap-frontend/src/containers/Training/Training.jsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useRef, useCallback } from "react";
+import { useState, useEffect, useRef, useCallback, useMemo } from "react";
 import { useParams } from "react-router-dom";
 import { Trash2 } from "lucide-react";
 import io from "socket.io-client";
@@ -40,6 +40,32 @@ import {
   trainingHistory as trainingHistoryAtom,
   modelListSelector,
 } from "../../shared/atoms";
+
+const PROBLEM_TYPE = Object.freeze({
+  BINARY_CLASSIFICATION: "1",
+  REGRESSION: "2",
+  MULTI_CLASS_CLASSIFICATION: "3",
+});
+
+const DEFAULT_LOSS_BY_PROBLEM_TYPE = {
+  [PROBLEM_TYPE.BINARY_CLASSIFICATION]: "sparse_categorical_crossentropy",
+  [PROBLEM_TYPE.MULTI_CLASS_CLASSIFICATION]: "sparse_categorical_crossentropy",
+  [PROBLEM_TYPE.REGRESSION]: "mean_squared_error",
+};
+
+// Cleaned up options (removed opaque 'key' property)
+const lossOptions = [
+  {
+    type: "classification",
+    label: "Sparse Categorical Crossentropy",
+    value: "sparse_categorical_crossentropy",
+  },
+  { type: "classification", label: "Categorical Crossentropy", value: "categorical_crossentropy" },
+  { type: "classification", label: "Binary Crossentropy", value: "binary_crossentropy" },
+  { type: "regression", label: "Mean Squared Error", value: "mean_squared_error" },
+  { type: "regression", label: "Mean Absolute Error", value: "mean_absolute_error" },
+  { type: "regression", label: "Huber", value: "huber" },
+];
 
 const optimizerOptions = [
   { key: "opt_1", label: "Adam", value: "adam" },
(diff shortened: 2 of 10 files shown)


tensormap-backend/app/schemas/deep_learning.py
```diff
@@ -97,4 +107,5 @@ class TrainingConfigRequest(BaseModel):
     optimizer: str = Field(min_length=1)
     metric: str = Field(min_length=1)
     epochs: int = Field(gt=0)
-    project_id: uuid_pkg.UUID | None = None
+    loss: LossFunction
+    project_id: uuid_pkg.UUID | None = None
```
ivantha: **[LOW]** Missing newline at end of file

The diff explicitly shows `\ No newline at end of file` at the last line of `deep_learning.py`. This violates the POSIX standard (text files must end with a newline), causes noisy git diffs in future PRs, and will trigger warnings in most linters (flake8 `W292`, pre-commit `end-of-file-fixer`, etc.).

**Suggestion:**
```
Add a single trailing newline character after the last line of the file. Most editors have a setting to do this automatically (`"files.insertFinalNewline": true` in VS Code).
```

tensormap-frontend/src/containers/Training/Training.jsx
```diff
@@ -340,6 +363,36 @@ export default function Training() {
...
+                  <SelectContent>
+                    <SelectGroup>
+                      <SelectLabel className="font-bold text-slate-500">Classification</SelectLabel>
+                      {lossOptions.filter(o => o.type === "classification").map((o) => (
+                        <SelectItem key={o.key} value={o.value}>
+                          {o.label}
+                        </SelectItem>
+                      ))}
+                    </SelectGroup>
+                    <SelectGroup>
```
ivantha: **[LOW]** Inline `Array.filter()` on a static constant called on every render

`lossOptions.filter(o => o.type === "classification")` and `lossOptions.filter(o => o.type === "regression")` are called inline inside JSX on every render. Since `lossOptions` is a static module-level constant that never changes, these filter operations produce the same result every single time and represent unnecessary computation per render cycle.

**Suggestion:**
```
Pre-compute the filtered arrays at module scope alongside the source array — no `useMemo` needed since neither array depends on component state:

```js
const lossOptions = [ /* ... */ ];
const classificationLossOptions = lossOptions.filter(o => o.type === "classification");
const regressionLossOptions = lossOptions.filter(o => o.type === "regression");

// In JSX:
{classificationLossOptions.map(o => <SelectItem key={o.value} value={o.value}>{o.label}</SelectItem>)}
{regressionLossOptions.map(o => <SelectItem key={o.value} value={o.value}>{o.label}</SelectItem>)}
```
```

File: tensormap-frontend/src/containers/Training/Training.jsx [hunk 3]
```diff
@@ -26,6 +28,15 @@ import {
 import { getAllFiles } from "../../services/FileServices";
 import { models as modelListAtom } from "../../shared/atoms";
 
+const lossOptions = [
+  { type: "classification", key: "loss_1", label: "Sparse Categorical Crossentropy", value: "sparse_categorical_crossentropy" },
+  { type: "classification", key: "loss_2", label: "Categorical Crossentropy", value: "categorical_crossentropy" },
+  { type: "classification", key: "loss_3", label: "Binary Crossentropy", value: "binary_crossentropy" },
+  { type: "regression", key: "loss_4", label: "Mean Squared Error", value: "mean_squared_error" },
+  { type: "regression", key: "loss_5", label: "Mean Absolute Error", value: "mean_absolute_error" },
+  { type: "regression", key: "loss_6", label: "Huber", value: "huber" },
```
ivantha: **[LOW]** `key` props use opaque sequential strings (`"loss_1"`) instead of the semantic `value`

The `lossOptions` array uses keys like `"loss_1"`, `"loss_2"`, ..., `"loss_6"` as React list keys. These are opaque, sequential identifiers with no semantic meaning. The `value` field (`"sparse_categorical_crossentropy"`, `"huber"`, etc.) is already a unique, stable, and meaningful identifier for each option — it should be used as the key instead.

**Suggestion:**
```
Remove the `key` field from `lossOptions` objects and use `value` directly as the React key:

```js
const lossOptions = [
  { type: "classification", label: "Sparse Categorical Crossentropy", value: "sparse_categorical_crossentropy" },
  { type: "classification", label: "Categorical Crossentropy",        value: "categorical_crossentropy" },
  // ...
];

// In JSX:
{classificationLossOptions.map(o => (
  <SelectItem key={o.value} value={o.value}>{o.label}</SelectItem>
))}
```
```

ivantha (commented):
## ◎ Argus Review

This PR makes a solid architectural improvement by replacing hardcoded binary loss logic with a structured mapping system backed by Pydantic `Literal` validation — the schema approach is idiomatic and well-conceived. However, there are significant correctness and compatibility issues that need to be resolved before merging: the `from_logits=True` assumption is hardcoded for all cross-entropy losses and will silently produce incorrect training results for models with softmax output layers, and the new required `loss` field on both schemas is a breaking change that will break existing API clients and database-stored configs without a migration or default. The frontend auto-select UX is a nice touch, but the magic numbers for problem type IDs and the unfiltered loss dropdown introduce fragility.

ivantha:

NamanmeetSingh:
I have successfully rebased the branch with the latest main. All merge conflicts in the backend and frontend have been resolved.

ivantha:

PR #161, opened 2026-02-26
Fix: Improve model validation error messages (#146)

### LINKED ISSUE #146: [Feature] Explain shape mismatches and bad params in validation errors
**Difficulty:** Medium
**Area:** backend

## Description
Already landed: graph-topology errors now produce friendly text in `tensormap-backend/app/services/model_generation.py` (no input layer at :41, unreachable nodes named at :84, no output layer at :100, dangling input at :109), surfaced as `{"success": false, "message": ...}` by the `except ValueError` at `tensormap-backend/app/services/deep_learning.py:163-164`. Three failure modes are still cryptic. Shape mismatches never reach that handler; they escape `model_from_json` into the generic `except Exception` at deep_learning.py:169-174, which either matches a substring in `app/shared/errors.py` (best case "Error in Model values") or returns "Model validation failed. Please recheck the model and inputs". An invalid parameter raises a bare `ValueError` out of `int(params["units"])` at model_generation.py:163 with no node or parameter name attached. An unrecognised node type raises `Unknown node type: {node_type}` (model_generation.py:200, mirrored by `Unknown layer type: {layer_type}` at `app/generators/tensorflow_generator.py:179`) without listing what is supported.

## Acceptance Criteria
- [ ] A shape mismatch between adjacent layers returns a message naming the two nodes and, where a rank change is the cause, saying that a Flatten layer is needed between them
- [ ] Parameter coercion failures name the node id and the parameter (for example `units` on `dense_2`) and state the value received, instead of surfacing a bare `int()` or `float()` `ValueError`
- [ ] `Unknown node type` and `Unknown layer type` errors list the supported types, sourced from `LAYER_REGISTRY` rather than a hardcoded list that can drift from it
- [ ] No message exposes a file path, a stack trace, or a TensorFlow internal class name
- [ ] Both generator paths are covered: the registry-driven `TensorFlowGenerator` and the legacy `model_generation` fallback that runs when it fails
- [ ] Valid models still validate, with the response shape unchanged
- [ ] Tests in `tensormap-backend/tests/test_model_generation.py` cover one shape mismatch, one invalid parameter value, and one unknown node type

## Resources
- `tensormap-backend/app/services/model_generation.py:147-200` - legacy `_build_layer`, including `int(params["units"])` at :163 and the `Unknown node type` raise at :200
- `tensormap-backend/app/generators/tensorflow_generator.py:159-199` - registry-driven `_build_layer`, with `Unknown layer type` at :179 and a catch-all wrap at :198-199 that swallows the real cause into one string
- `tensormap-backend/app/services/deep_learning.py:160-174` - where generator errors become API responses; the friendly path is the `except ValueError` at :163, the cryptic path is the `except Exception` at :169
- `tensormap-backend/app/shared/errors.py:3-10` - the substring-to-message table the generic handler falls back on
- `tensormap-backend/app/layers/registry.py:453` - `get_layer_spec`, which already formats an "Available types" list that the generator errors could reuse
- `tensormap-backend/tests/test_model_generation.py:393-460` - existing coverage for the topology messages that already landed
- [TensorFlow error types](https://www.tensorflow.org/api_docs/python/tf/errors)

kanaksharma9:
## Description

This PR improves the error handling in `model_generation.py` to catch specific failure modes and return user-friendly error messages, as requested in the issue. 

Specific changes:
* Added a pre-traversal check to catch missing `custominput` layers.
* Implemented post-traversal validation to detect disconnected nodes in the graph.
* Added exception handling in `_build_layer()` to catch type/parameter errors for Dense and Conv2D nodes.
* Added specific exception handling for unknown node types and TensorFlow shape mismatches.
* Wrapped the `model_validate_service` in `routers/deep_learning.py` with a `try/except` block to return the required `{"success": false, "message": "..."}` JSON format and prevent 500 errors from leaking stack traces.

Fixes #146

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## How Has This Been Tested?

I spun up the PostgreSQL database via Docker, ran the Alembic migrations and started the FastAPI server locally. I then manually sent invalid graph payloads (missing inputs, disconnected graphs, etc.) to the `/model/validate` endpoint to verify that the server returned a `400` status code with the correctly formatted `{"success": false, "message": "..."}` JSON payload instead of crashing.

- [ ] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Checklist

- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/tensormap-backend/app/routers/deep_learning.py
+++ b/tensormap-backend/app/routers/deep_learning.py
@@ -28,8 +28,16 @@ router = APIRouter(tags=["deep-learning"])
 async def validate_model(request: ModelValidateRequest, db: Session = Depends(get_db)):
     """Validate a ReactFlow graph as a Keras model and persist the configuration."""
     logger.debug("Validating model for project_id=%s", request.project_id)
-    body, status_code = model_validate_service(db, incoming=request.model_dump(), project_id=request.project_id)
-    return JSONResponse(status_code=status_code, content=body)
+
+    try:
+        body, status_code = model_validate_service(db, incoming=request.model_dump(), project_id=request.project_id)
+        return JSONResponse(status_code=status_code, content=body)
+
+    except ValueError as e:
+        return JSONResponse(status_code=400, content={"success": False, "message": str(e)})
+    except Exception as e:
+        logger.error("Unexpected error during model validation: %s", e, exc_info=True)
+        return JSONResponse(status_code=500, content={"success": False, "message": "Internal Error"})
 
 
 @router.post("/model/save")
--- a/tensormap-backend/app/services/model_generation.py
+++ b/tensormap-backend/app/services/model_generation.py
@@ -10,6 +10,21 @@ from app.shared.logging_config import get_logger
 logger = get_logger(__name__)
 
 
+def _parse_strict_int(value, param_name: str, node_name: str) -> int:
+    """Parse an integer while rejecting lossy coercions such as float -> int."""
+    if isinstance(value, bool):
+        raise ValueError(f"Invalid '{param_name}' parameter for node '{node_name}': expected an integer, got '{value}'")
+    if isinstance(value, int):
+        return value
+    if isinstance(value, str):
+        normalized = value.strip()
+        if normalized.startswith("-"):
+            normalized = normalized[1:]
+        if normalized.isdigit():
+            return int(value)
+    raise ValueError(f"Invalid '{param_name}' parameter for node '{node_name}': expected an integer, got '{value}'")
+
+
 def model_generation(model_params: dict) -> dict:
     """Transform ReactFlow nodes and edges into a Keras functional-API JSON structure.
 
@@ -32,6 +47,9 @@ def model_generation(model_params: dict) -> dict:
     keras_tensors = {}
     visited = set()
     queue = []
+
+    if not any(n["type"] == "custominput" for n in model_params["nodes"]):
+        raise ValueError("No Input layer found. Please add an Input node to start the network.")
 
     for node in model_params["nodes"]:
         if node["type"] == "custominput":
@@ -64,6 +82,13 @@ def model_generation(model_params: dict) -> dict:
             node = nodes_by_id[target_id]
             keras_tensors[target_id] = _build_layer(node, input_tensor)
 
+    if len(visited) < len(model_params["nodes"]):
+        unconnected_count = len(model_params["nodes"]) - len(visited)
+        raise ValueError(
+            f"Disconnected graph: {unconnected_count} node(s) are not connected "
+            "to the Input layer. Please ensure all layers are linked."
+        )
+
     # Identify input and output tensors
     inputs = [keras_tensors[n["id"]] for n in model_params["nodes"] if n["type"] == "custominput"]
     output_ids = [n["id"] for n in model_params["nodes"] if n["id"] not in source_to_targets]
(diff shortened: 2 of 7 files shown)


ivantha (changes requested):
Bug: Disconnected graph check is inside the while loop (model_generation.py:72-74)

  It runs after processing each node's neighbors, not after BFS completes — so it'll false-positive on any valid
  multi-layer graph. Needs to be dedented one level to run after the loop exits.

  README broken

  tensormap-backend/README.md:110 — the pytest command was accidentally replaced with http://localhost:4300/docs.

  Ruff linting fails (12 errors)

  Run uv run ruff check on the changed files. Main issues: lines over 120 chars, unused variables (units, filters,
  etc. are validated then recomputed instead of reused), and missing raise ... from e in except blocks (B904).

  Minor: Inconsistent indentation on lines 112/130, and package-lock.json changes look unrelated.

ivantha (commented):
## ◎ Argus Review

This PR has good intentions — improving error messages for model validation failures — but contains a **critical bug** in the disconnected graph check that would cause `ValueError` to be raised for *every* valid graph, because `custominput` nodes are never added to the `visited` set used in the comparison. This regression would break all model validation. Beyond that bug, there are structural issues in `_build_layer` (fragile string-matching for TF error detection, `raise e` destroying tracebacks, overly broad nested try/except), missing `exc_info=True` on the error logger, and a complete absence of automated tests. The sanitization of 500-error responses and use of `raise ... from e` exception chaining are positive patterns worth keeping — but the critical correctness issue must be resolved before merging.

### ⚠️ Breaking Changes

- **[INFO]** Behavioral Change: model_generation Now Raises ValueError Where It Previously May Have Raised Other Exception Types: **Previous behavior**: `model_generation()` had no explicit validation for missing input nodes or disconnected graphs. Invalid parameters to `Dense`/`Conv2D` would surface as raw `ValueError`, `KeyError`, `TypeError`, or TensorFlow-internal exceptions — potentially exposing stack traces to callers.

**New behavior**: Several failure modes now consistently raise `ValueError` with user-friendly messages. The router explicitly catches `ValueError → 400` and `Exception → 500` sanitized response.

**Who is affected**: Any code (tests, scripts, other services) that calls `model_generation()` or `model_validate_service()` directly and catches specific exception types (e.g., `KeyError`, `TypeError`) or depends on uncaught exceptions propagating will be affected.

**Migration path**: Update callers to handle `ValueError` for validation failures. The new structured error messages are stable and user-readable.
  > **Suggestion:** Document the new exception contract of `model_generation()` with a docstring listing what `ValueError`s it can raise and under what conditions.


### PR Description Issues

- **[LOW]** PR Checklist: Neither Existing Tests Passing Nor New Tests Are Confirmed: The PR checklist shows:
```
- [ ] Existing tests pass
- [ ] New tests added
```
Both are unchecked. Given that this PR modifies the core validation logic in `model_generation.py`, this is a meaningful gap. If existing tests pass, that checkbox should be checked. If they don't (or don't exist), that should be called out explicitly as a known risk or follow-up item.

The omission of new tests is especially concerning given the new validation branches introduced — one of which (the disconnected graph check) contains a critical false-positive bug that unit tests would have caught.
  > **Suggestion:** Update the PR description to explicitly state the test situation:
```
- [ ] Existing tests pass ← State whether tests exist and their status
- [ ] New tests added ← Acknowledge this as a follow-up or add a linked issue
```
Or add a "Known Limitations" section: _"No automated tests currently cover model generation validation; a follow-up issue (#XXX) has been filed to add test coverage."_
- **[LOW]** PR Description Omits Risk/Rollback Section for Logic Changes in Core Service: The PR modifies `model_generation.py`, which is the core service powering model validation. The description describes what was changed but doesn't include:
- Any mention of potential regressions in the happy path
- Whether the endpoint was tested with previously-valid payloads (not just error cases)
- A rollback plan or feature flag if the new validation is too strict

Given the critical bug found (disconnected graph check fires on all valid graphs), this omission means reviewers had no prompt to verify the happy path was tested.
  > **Suggestion:** Add a **Happy Path Testing** bullet under "How Has This Been Tested?":
> Verified that a valid graph (input → flatten → dense) still returns `{"success": true}` with status 200 after the changes.

ivantha:

kanaksharma9:
@ivantha  All checks passing now, rebased onto latest main, squashed to a single commit and fixed all review findings:
Disconnected-graph bug (input nodes now seeded into visited)
raise e → bare raise to preserve tracebacks
_build_layer refactored: param parsing split from TF call
exc_info=True added to logger
Strict int parser to reject float units/filters
7 unit tests added covering happy path and all error branches
Frontend lint and Prettier clean
Ready for re-review.

ivantha:
Thank you for improving the model validation error messages (#146)! This PR has been superseded by #205 which addresses the same issue, is conflict-free, and is ready for review. We appreciate your work on making these errors more user-friendly.

PR #166, opened 2026-02-26
feat: Add canvas connection validation with toast notifications

### LINKED ISSUE #126: [Feature] Reject invalid canvas connections with inline feedback
**Difficulty:** Medium
**Area:** frontend

## Description
`onConnect` in `tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx:329-335` calls `addEdge(params, eds)` unconditionally, and the `<ReactFlow>` element at :704-723 passes no `isValidConnection`, so self-loops, duplicate edges and extra incoming edges on an already-fed node are all accepted. The user only finds out at validate or train time, as a backend error. This issue adds an `isValidConnection` guard plus a short-lived inline message on the canvas explaining why an edge was refused. The rules should build on what the backend already enforces rather than invent a second rule set: `LAYER_REGISTRY` in `tensormap-backend/app/layers/registry.py:108` is the source of truth for layer types and which ones accept multiple inputs, `validate_ir_graph` in `tensormap-backend/app/ir/schema.py:312` covers single-input, Concatenate arity and the DAG constraint, and `POST /layers/validate-graph` exposes that check to the frontend.

## Acceptance Criteria
- [ ] `isValidConnection` is passed to `<ReactFlow>` so an invalid drag is marked before the user drops it
- [ ] Self-loops (source node id equal to target node id) are refused
- [ ] A duplicate edge between the same source and target handles is refused
- [ ] A second incoming edge on a target is refused, except for layer types that legitimately merge inputs (Concatenate, per `LAYER_REGISTRY` at registry.py:434)
- [ ] Each refusal shows a short message near the canvas naming the reason, auto-dismissing after about 3 seconds, and is not a browser `alert()`
- [ ] Valid connections still add the edge and still push an undo snapshot via `takeSnapshotAndUpdate`
- [ ] Deleting or reconnecting existing edges is unaffected
- [ ] Tests in `tensormap-frontend/src/components/DragAndDropCanvas/Canvas.test.jsx` cover each refused case plus one accepted connection

## Resources
- `tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx:329-335` - `onConnect`, which currently adds any dragged edge
- `tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx:704-723` - the `<ReactFlow>` prop list, with no `isValidConnection`
- `tensormap-frontend/src/components/DragAndDropCanvas/Helpers.ts:99` - `isGraphConnected`, with `canSaveModel` at :151 and `canSaveModelSimple` at :234; the file is now TypeScript, not `Helpers.jsx`
- `tensormap-frontend/src/hooks/useLayerRegistry.ts:101` - `useLayerRegistry`, already called by the canvas, plus `getLayerSpec` at :136 for per-type metadata
- `tensormap-backend/app/layers/registry.py:108` - `LAYER_REGISTRY`, including the `merge` flag documented at :88
- `tensormap-backend/app/routers/layers.py:26` - `POST /layers/validate-graph`, which runs `validate_ir_graph` and returns `{valid, errors}`
- `tensormap-frontend/src/components/ui/` - existing shadcn primitives; there is no toast component or toast dependency in the project yet, so the feedback surface has to be built or added
- [ReactFlow isValidConnection](https://reactflow.dev/api-reference/react-flow#is-valid-connection)
- [ReactFlow onConnect](https://reactflow.dev/api-reference/react-flow#on-connect)

anudeeps0306:
## Description

This PR adds a robust **connection validation system** to the ReactFlow canvas along with a **self-contained toast notification** to clearly communicate errors to users when invalid connections are attempted.

### Toast Notification

* Added a reusable toast notification that:

  * Appears at the **top-center of the canvas**
  * Uses a **red warning style**
  * **Auto-dismisses after 3 seconds**
  * Includes a **manual close button**
* The toast is fully self-contained and does not block canvas interactions.

### Connection Validation (`Canvas.jsx`)

Enhanced ReactFlow’s `isValidConnection` callback (which runs on every hover during a drag) to **reject invalid connections** and store the rejection reason in a ref. The following cases are now prevented:

* Same handle type connections (**source → source** or **target → target**)
* **Self-loops** (connecting a node to itself)
* **Multiple incoming edges** on the same target
* **Duplicate edges**
* **Cycles** (if `A → B` already exists, `B → A` is rejected using a BFS traversal)

### Why `onConnectEnd` Alone Was Not Enough

ReactFlow **silently swallows invalid drops** (e.g., dropping on an occupied handle), so `onConnectEnd` is not reliably triggered for all invalid cases.

To solve this, we added a **global `window` mouseup listener** that:

* Checks whether a connection drag was in progress (tracked via `connectionDragRef`, set in `onConnectStart` and cleared in `onConnect` for successful connections)
* Displays the stored error message from `connectionErrorRef` if present
* As a fallback, walks the DOM using ReactFlow’s own `data-handleid` / `data-nodeid` attributes to detect drops on occupied handles and show the appropriate error

This guarantees users always get clear feedback when a connection attempt fails.

### Validation

* Manually tested all invalid connection scenarios:

  * Same-handle connections
  * Self-loops
  * Duplicate edges
  * Multiple incoming edges
  * Cycles
* Verified toast appears with correct message and:

  * Auto-dismisses after 3 seconds
  * Can be closed manually
* Confirmed valid connections still work normally without interference
* Built the frontend successfully with `npm run build` (no errors or warnings)

Fixes #126

---

## Type of Change

* [ ] Bug fix
* [x] New feature
* [ ] Breaking change
* [ ] Documentation update

---

## How Has This Been Tested?

* Manually tested all invalid connection cases to confirm they are rejected with the correct error message

* Verified toast behavior (position, auto-dismiss, manual close)

* Confirmed valid connections still succeed

* Ran `npm run build` to ensure production build passes

* [x] Existing tests pass

* [ ] New tests added

* [x] Manual testing

---

image
image

---

## Checklist

* [x] My code follows the project's style guidelines
* [x] I have performed a self-review
* [ ] I have added/updated documentation as needed
* [x] My changes generate no new warnings
* [x] Tests pass locally

---


rajutkarsh07:
LGTM just rebase onto current main, and fix the lint error

tensormap-frontend/src/components/DragAndDropCanvas/Toast.jsx
```diff
@@ -0,0 +1,24 @@
...
+
+export default function Toast({ message, onClose }) {
+  useEffect(() => {
+    if (message) {
+      const timer = setTimeout(() => {
+        onClose();
+      }, 3000);
+      return () => clearTimeout(timer);
+    }
+  }, [message, onClose]);
```
ivantha: **[HIGH]** Toast auto-dismiss timer resets on every Canvas render due to unstable `onClose` reference

The `Toast` component's `useEffect` has `[message, onClose]` as dependencies. In `Canvas.jsx`, `onClose` is passed as an inline arrow function `() => setToastMessage(null)`, which creates a **new function reference on every render**. Any state change in `Canvas` (e.g., a node being moved, an edge being added) triggers a re-render, which creates a new `onClose` reference, which restarts the 3-second timer. In a dynamic canvas environment where renders are frequent, the toast may **never auto-dismiss**. This is a functional defect that directly undermines the advertised 3-second auto-dismiss feature.

**Suggestion:**
```
Either memoize `onClose` in Canvas.jsx with `useCallback`:
```jsx
const handleToastClose = useCallback(() => setToastMessage(null), []);
// ...
<Toast message={toastMessage} onClose={handleToastClose} />
```
Or, inside `Toast.jsx`, track `onClose` via a ref so the effect only re-runs when `message` changes:
```jsx
export default function Toast({ message, onClose }) {
  const onCloseRef = useRef(onClose);
  useEffect(() => { onCloseRef.current = onClose; });

  useEffect(() => {
    if (message) {
      const timer = setTimeout(() => onCloseRef.current(), 3000);
      return () => clearTimeout(timer);
    }
  }, [message]); // stable: only message drives the timer
  // ...
}
```
```

File: tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx [hunk 2]
```diff
@@ -192,7 +195,143 @@ function Canvas() {
     setModelName("");
   }, [draftKey, setNodes, setEdges]);
 
-  const onConnect = useCallback((params) => setEdges((eds) => addEdge(params, eds)), [setEdges]);
+  const onConnect = useCallback(
+    (params) => {
+      // Mark connection as successful so mouseup doesn't show a false error
+      connectionDragRef.current = null;
+      setEdges((eds) => addEdge(params, eds));
+    },
```
ivantha: **[MEDIUM]** `connectionDragRef` is used before it is declared

Inside the `onConnect` callback (lines 196–203), `connectionDragRef.current = null` is written. However, `connectionDragRef` is declared via `useRef` at line 260 — *after* `onConnect`. While this works at runtime because `onConnect` is only invoked asynchronously (after render completes, by which time `connectionDragRef` is fully initialized), it violates the **temporal dead zone** rule for `const` in JavaScript and is a significant code clarity issue. Any reader will naturally wonder "how does this work?" and the linter may flag it as a potential error in strict mode analysis tools.

**Suggestion:**
```
Move the `connectionDragRef` declaration to before the `onConnect` callback (and before `connectionErrorRef` too, since it is used in `onConnectStart`):
```jsx
// Declare refs before any callback that references them
const connectionErrorRef = useRef(null);
const connectionDragRef = useRef(null);

const onConnect = useCallback((params) => {
  connectionDragRef.current = null;
  setEdges((eds) => addEdge(params, eds));
}, [setEdges]);
```
```

File: tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx [hunk 3]
```diff
@@ -192,7 +195,143 @@ function Canvas() {
...
+        connectionErrorRef.current = "Target node already has an incoming connection";
+        return false;
+      }
+
+      // Prevent duplicate connections
+      const isDuplicate = edges.some(
+        (e) => e.source === connection.source && e.target === connection.target,
+      );
+      if (isDuplicate) {
+        connectionErrorRef.current = "Connection already exists";
```
ivantha: **[MEDIUM]** Duplicate-edge check in `isValidConnection` is unreachable dead code

The `hasIncoming` check (lines 222–227) returns `false` for **any** connection whose target already has an incoming edge — regardless of source. Since a duplicate edge `(A→B)` means B already has at least one incoming edge (from A), the `hasIncoming` guard will always fire first and return before reaching the `isDuplicate` check. The `isDuplicate` block is therefore **dead code that can never be executed** within `isValidConnection`.

This is not a functional bug per se (duplicates are still blocked), but it indicates a logical ordering mistake and leaves misleading code in the codebase.

**Suggestion:**
```
Remove the duplicate check from `isValidConnection` since it is already subsumed by the `hasIncoming` check. The duplicate check remains valid and useful in the `handleMouseUp` fallback (where source-specific matching is needed), so keep it there. Alternatively, if the intent is to distinguish these two errors, the `isDuplicate` check must come *before* `hasIncoming`:
```js
// Check duplicate first (more specific error message)
const isDuplicate = edges.some(
  (e) => e.source === connection.source && e.target === connection.target,
);
if (isDuplicate) {
  connectionErrorRef.current = "Connection already exists";
  return false;
}

// Then check multiple incoming (covers non-duplicate case)
const hasIncoming = edges.some((e) => e.target === connection.target);
if (hasIncoming) {
  connectionErrorRef.current = "Target node already has an incoming connection";
  return false;
}
```
```

File: tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx [hunk 4]
```diff
@@ -192,7 +195,143 @@ function Canvas() {
...
+    [edges],
+  );
+
+  // Store the source node/handle when dragging starts so we can validate on mouseup
+  const connectionDragRef = useRef(null);
+
+  const onConnectStart = useCallback((_event, { nodeId, handleId, handleType }) => {
+    connectionErrorRef.current = null;
+    connectionDragRef.current = { nodeId, handleId, handleType };
+  }, []);
```
ivantha: **[MEDIUM]** Escape key during connection drag leaves `connectionDragRef` in a stale non-null state

`onConnectEnd` is defined as an empty no-op callback. In ReactFlow, pressing Escape to cancel a connection drag triggers `onConnectEnd` but does **not** trigger `onConnect`. Since neither `onConnect` (only called on success) nor the no-op `onConnectEnd` clears `connectionDragRef`, it remains set after an Escape cancellation.

Consequence: The next `mouseup` event (e.g., a simple canvas click after Escape) will find a non-null `drag` in `handleMouseUp`. If the click happens near a ReactFlow handle, the DOM walk may identify it, and the fallback duplicate/incoming-edge checks could fire and display a **completely spurious toast notification**.

**Suggestion:**
```
Use `onConnectEnd` to clear the drag ref rather than the global mouseup listener:
```js
const onConnectEnd = useCallback(() => {
  // Clear drag state when ReactFlow signals the connection attempt is over
  // (covers Escape, drop on canvas, and all non-success endings)
  connectionDragRef.current = null;
  connectionErrorRef.current = null;
}, []);
```
With this, the global `mouseup` fallback can focus purely on the "silent swallow" cases that `onConnectEnd` misses, and the stale-ref issue is eliminated.
```

File: tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx [hunk 5]
```diff
@@ -192,7 +195,143 @@ function Canvas() {
...
+      let el = event.target;
+      let targetHandleId = null;
+      let targetNodeId = null;
+      while (el && el !== document.body) {
+        const hid = el.getAttribute && el.getAttribute("data-handleid");
+        if (hid !== null && hid !== undefined) {
+          targetHandleId = hid;
+          targetNodeId = el.getAttribute("data-nodeid");
+          break;
+        }
```
ivantha: **[MEDIUM]** Global `window` mouseup listener relies on undocumented ReactFlow internal DOM attributes

The fallback validation mechanism walks the DOM using `data-handleid` and `data-nodeid` attributes. The comment acknowledges these are from "ReactFlow v11". These are **internal implementation details** of the ReactFlow library — they are not part of its public API or documented contract and can change in any minor or patch release.

Additionally, `window.addEventListener('mouseup', ...)` is a global side effect that:
- Runs for every mouseup on the entire page (not scoped to the canvas)
- May conflict with other mouseup handlers, including those inside ReactFlow itself
- Re-registers on every `edges` state change (every time an edge is added/removed), though cleanup is correct

This approach is fragile by design and will likely break on a ReactFlow upgrade.

**Suggestion:**
```
Consider using ReactFlow's `useOnViewportChange` or the `onPaneClick` prop for canvas drop detection. A more robust approach would be to use ReactFlow's public APIs:
- Use `onConnectEnd` (with the `event` parameter it receives) for detecting drop location
- Use ReactFlow's `useReactFlow()` hook methods like `getIntersectingNodes` for position-based logic

If the global mouseup approach must be kept (e.g., because `onConnectEnd` truly doesn't fire in some cases), pin the ReactFlow version in `package.json` and add a code comment explicitly noting the version dependency and the risk of breakage.
```

tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx
```diff
@@ -192,7 +195,143 @@ function Canvas() {
...
+      // Target already has an incoming edge
+      const hasIncoming = edges.some((e) => e.target === targetNodeId);
+      if (hasIncoming) {
+        setToastMessage("Target node already has an incoming connection");
+      }
+    };
+
+    window.addEventListener("mouseup", handleMouseUp);
+    return () => {
+      window.removeEventListener("mouseup", handleMouseUp);
```
ivantha: **[MEDIUM]** Cycle detection is missing from the `handleMouseUp` fallback validation path

The `isValidConnection` callback performs BFS cycle detection (lines 238–251). However, the `handleMouseUp` fallback handler — which is invoked when ReactFlow "silently swallows" a connection drop — only checks for self-loops, wrong direction, duplicate edges, and multiple incoming edges. **Cycle detection is absent from the fallback.**

If ReactFlow swallows a cyclic connection attempt (per the PR's stated rationale for the global mouseup approach), the user will receive **no feedback** that a cycle was rejected. The PR's own justification makes this gap inconsistent with the feature's stated goal.

**Suggestion:**
```
Extract the BFS logic into a shared utility function and call it from both `isValidConnection` and `handleMouseUp`:
```js
// Outside the component, or in a helpers file
function wouldCreateCycle(edges, sourceId, targetId) {
  const visited = new Set();
  const queue = [targetId];
  while (queue.length > 0) {
    const curr = queue.shift();
    if (curr === sourceId) return true;
    if (visited.has(curr)) continue;
    visited.add(curr);
    edges.filter((e) => e.source === curr).forEach((e) => queue.push(e.target));
  }
  return false;
}

// In handleMouseUp fallback:
if (wouldCreateCycle(edges, drag.nodeId, targetNodeId)) {
  setToastMessage("This connection would create a cycle");
  return;
}
```
```

File: tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx [hunk 7]
```diff
@@ -192,7 +195,143 @@ function Canvas() {
...
+      // source→target would form a cycle.
+      const visited = new Set();
+      const queue = [connection.target];
+      while (queue.length > 0) {
+        const curr = queue.shift();
+        if (curr === connection.source) {
+          connectionErrorRef.current = "This connection would create a cycle";
+          return false;
+        }
+        if (visited.has(curr)) continue;
```
ivantha: **[LOW]** O(V+E) BFS cycle detection runs synchronously on every handle hover during a drag

The `isValidConnection` callback, which includes a full BFS traversal of the edge graph, is called by ReactFlow **on every pointer-enter event over a potential target handle** during a connection drag. For a canvas with many nodes and edges, this synchronous O(V+E) computation on each hover event can cause noticeable jank, especially as it also involves `edges.filter().forEach()` allocations inside the loop.

Additionally, `edges.filter((e) => e.source === curr)` is called for every node visited in the BFS, resulting in O(V × E) in the worst case without an adjacency map.

**Suggestion:**
```
Pre-build an adjacency list from `edges` (a `Map<nodeId, nodeId[]>`) using `useMemo`, and reuse it in the BFS:
```js
const adjacency = useMemo(() => {
  const map = new Map();
  edges.forEach((e) => {
    if (!map.has(e.source)) map.set(e.source, []);
    map.get(e.source).push(e.target);
  });
  return map;
}, [edges]);

// In BFS:
(adjacency.get(curr) ?? []).forEach((t) => queue.push(t));
```
This reduces BFS from O(V×E) to O(V+E) and avoids repeated array allocations on every hover.
```

File: tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx [hunk 8]
```diff
@@ -192,7 +195,143 @@ function Canvas() {
...
+
+      if (!targetHandleId || !targetNodeId) return;
+      if (drag.handleType !== "source") return;
+
+      // Self-loop (belt-and-suspenders — also caught by isValidConnection)
+      if (targetNodeId === drag.nodeId) {
+        setToastMessage("Cannot connect a node to itself");
+        return;
+      }
+
```
ivantha: **[LOW]** Wrong-direction handle drop silently fails in `handleMouseUp` fallback without any toast

In `handleMouseUp`, if the user drops a source handle onto another source handle (source→source), the code reaches the direction check:
```js
if (!drag.handleId?.endsWith('_out') || !targetHandleId.endsWith('_in')) return;
```
This `return` exits silently with no toast. The `isValidConnection` callback *should* have caught this and stored an error in `connectionErrorRef`, but if `connectionErrorRef` was cleared (e.g., by the user hovering over a valid handle just before releasing), the user gets no feedback for a source→source drop that ReactFlow swallowed.

**Suggestion:**
```
Show a toast for invalid direction instead of silently returning:
```js
if (!drag.handleId?.endsWith('_out') || !targetHandleId.endsWith('_in')) {
  setToastMessage('Cannot connect handles of the same type');
  return;
}
```
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR introduces a meaningful and well-thought-out connection validation system for the ReactFlow canvas, tackling a real UX gap (silent failure on invalid connections). The cycle detection logic is algorithmically sound, the Toast component is cleanly isolated, and the dual-path validation strategy (isValidConnection + mouseup fallback) shows awareness of ReactFlow's behavioral quirks. However, there are two high-severity functional bugs: (1) the Toast timer resets on every Canvas render due to an unstable `onClose` reference, meaning the auto-dismiss feature may never fire in practice, and (2) a false-positive toast appears when a user hovers over an invalid handle but releases on empty canvas. Additionally, there is a medium-severity dead code issue (the duplicate-edge check in `isValidConnection` is unreachable), the global `window` mouseup listener is fragile due to reliance on ReactFlow's internal DOM attributes, and cycle detection is missing from the mouseup fallback path. The absence of automated tests for this complex, stateful logic is a notable gap given the number of subtle race conditions and edge cases involved.

File: tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx [hunk 9]
```diff
@@ -192,7 +195,143 @@ function Canvas() {
...
+      // Only valid drop direction: _out → _in
+      if (!drag.handleId?.endsWith("_out") || !targetHandleId.endsWith("_in")) return;
+
+      // Duplicate edge
+      const isDuplicate = edges.some((e) => e.source === drag.nodeId && e.target === targetNodeId);
```
ivantha: **[LOW]** Inconsistent error messages for the same validation rule across the two code paths

The `isValidConnection` callback and the `handleMouseUp` fallback handler use slightly different messages for the same semantic error:
- Duplicate edge: `"Connection already exists"` vs. `"Connection already exists between these nodes"`
- Self-loop: `"Cannot connect a node to itself"` (consistent ✓)
- Incoming edge: `"Target node already has an incoming connection"` (consistent ✓)

Users who encounter the duplicate-edge error via one path vs. the other will see different text, which is inconsistent and makes it harder to write documentation or tests.

**Suggestion:**
```
Define error messages as constants in a shared location (e.g., a `validationMessages.js` file or at the top of `Canvas.jsx`) and reference them from both code paths:
```js
const VALIDATION_MESSAGES = {
  sameHandleType: 'Cannot connect handles of the same type',
  selfLoop: 'Cannot connect a node to itself',
  multipleIncoming: 'Target node already has an incoming connection',
  duplicate: 'Connection already exists between these nodes',
  cycle: 'This connection would create a cycle',
};
```
```

ivantha:

ivantha:

PR #170, opened 2026-02-26
feat: add model deletion with cascade cleanup

### LINKED ISSUE #133: Model Deletion with Cascade Cleanup
**Difficulty:** Medium
**Area:** full stack

## Description

TensorMap currently has no way to delete a saved model. Over time, the model list accumulates old models that are no longer needed, cluttering the model selection dropdown and the database. Users need the ability to delete models they no longer want.

This issue adds a `DELETE /model/{id}` backend endpoint that removes a model and its associated `ModelConfigs` records (via database cascade), and a frontend delete button with a confirmation dialog on the model list or canvas page.

## Acceptance Criteria

- [ ] A delete endpoint exists for removing models by ID
- [ ] Deleting a model also removes its associated configuration records (cascade)
- [ ] The endpoint returns a proper error response if the model ID does not exist
- [ ] A delete button (with trash icon) is visible next to each model in the UI
- [ ] Clicking the delete button opens a confirmation dialog showing the model name
- [ ] Confirming deletion removes the model and refreshes the model list
- [ ] Canceling the dialog does not delete anything
- [ ] After deletion, if the deleted model was currently selected, the selection is cleared

## Resources

- `tensormap-backend/app/routers/deep_learning.py` — model endpoints
- `tensormap-backend/app/models/ml.py` — `ModelBasic` and `ModelConfigs` with relationship cascade
- `tensormap-backend/app/shared/constants.py` — `MODEL_GENERATION_LOCATION` for file cleanup
- `tensormap-frontend/src/services/ModelServices.js` — `getAllModels()` and model API functions
- `tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx` — model loading on mount
- `tensormap-frontend/src/components/ui/dialog.jsx` — shadcn/ui Dialog component
- `tensormap-frontend/src/shared/atoms.js` — `models` Recoil atom

kartikeyg0104:
Closes #133

## Summary

TensorMap previously had no way to delete saved models, causing the model list to accumulate stale entries over time.

This PR introduces a complete full-stack delete capability:
	•	Backend: DELETE /model/{id} endpoint with cascade cleanup
	•	Frontend: Per-model delete button with confirmation dialog in the Training UI

### Backend

**`tensormap-backend/app/services/deep_learning.py`**
- Added `delete_model_service(db, model_id)`: looks up `ModelBasic` by integer PK, deletes it (the existing `cascade="all,delete"` on the `configs` relationship automatically removes all associated `ModelConfigs` rows), and does a best-effort removal of the model JSON file from disk.
- Updated `get_available_model_list()` to return `{ id, model_name }` objects instead of bare strings, so the frontend can issue DELETE requests using the numeric DB id.

**`tensormap-backend/app/routers/deep_learning.py`**
- Added `DELETE /model/{model_id}` route backed by `delete_model_service`.

**`tensormap-backend/tests/test_deep_learning_service.py`** *(new)*
- 5 unit tests for `delete_model_service` using `MagicMock` — no live DB or TensorFlow required:
  - `test_success_deletes_model_and_file` — happy path, JSON file is removed
  - `test_model_not_found_returns_404` — returns 404 when id is unknown
  - `test_success_when_json_file_missing` — deletion still succeeds if JSON file is already gone
  - `test_db_error_rolls_back` — DB exception triggers rollback and returns 500
  - `test_get_called_with_correct_id` — verifies `db.get(ModelBasic, id)` call signature

**`tensormap-backend/tests/__init__.py`** *(new)* — makes `tests/` a proper Python package.

### Frontend

**`tensormap-frontend/src/constants/Urls.js`**
- Added `BACKEND_DELETE_MODEL = "/model"` constant.

**`tensormap-frontend/src/services/ModelServices.js`**
- Updated `getAllModels()` to handle the new `{ id, model_name }` response shape.
- Added `deleteModel(modelId)` — issues `DELETE /model/{modelId}`.

**`tensormap-frontend/src/containers/Training/Training.jsx`**
- Added `deleteConfirm` state, `handleDeleteClick`, and `handleDeleteConfirm` handlers.
- Renders a `<Trash2>` icon button next to each model in the dropdown.
- Clicking the trash icon opens a confirmation dialog showing the model name.
- Confirming calls `deleteModel`, removes the entry from the Recoil atom, and clears `selectedModel` if the deleted model was active.
- Cancelling closes the dialog with no side effects.

**`tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx`**
- Updated `loadModel()` to consume the new `{ id, model_name }` shape from `getAllModels()`.
- Replaced the optimistic `setModelList` push after save with a re-fetch of `getAllModels()` so every entry in the list carries a valid DB id (needed for deletion).

## Testing

```bash
# Backend unit tests (no DB / TF required)
cd tensormap-backend
uv run pytest tests/test_deep_learning_service.py -v
# 5 passed
```

### DIFF
--- a/tensormap-backend/app/routers/deep_learning.py
+++ b/tensormap-backend/app/routers/deep_learning.py
@@ -9,6 +9,7 @@ from sqlmodel import Session
 from app.database import get_db
 from app.schemas.deep_learning import ModelNameRequest, ModelSaveRequest, ModelValidateRequest, TrainingConfigRequest
 from app.services.deep_learning import (
+    delete_model_service,
     get_available_model_list,
     get_code_service,
     get_model_graph_service,
@@ -86,6 +87,17 @@ def get_model_graph(
 ):
     """Retrieve the full ReactFlow graph for a saved model."""
     body, status_code = get_model_graph_service(db, model_name=model_name, project_id=project_id)
+    return JSONResponse(status_code=status_code, content=body)
+
+
+@router.delete("/model/{model_id}")
+async def delete_model(
+    model_id: int,
+    db: Session = Depends(get_db),
+):
+    """Delete a saved model and its associated configuration records."""
+    logger.info("Deleting model id=%s", model_id)
+    body, status_code = delete_model_service(db, model_id=model_id)
     return JSONResponse(status_code=status_code, content=body)
 
 
--- a/tensormap-backend/app/services/deep_learning.py
+++ b/tensormap-backend/app/services/deep_learning.py
@@ -387,7 +387,31 @@ def get_available_model_list(
     total = db.exec(select(func.count()).select_from(base_filter.subquery())).one()
 
     models = db.exec(base_filter.offset(offset).limit(limit)).all()
-    data = [m.model_name for m in models]
+    data = [{"id": m.id, "model_name": m.model_name} for m in models]
     body = {"success": True, "message": "Model list generated successfully.", "data": data}
     body["pagination"] = {"total": total, "offset": offset, "limit": limit}
     return body, 200
+
+
+def delete_model_service(db: Session, model_id: int) -> tuple:
+    """Delete a model and its associated ModelConfigs (cascade), and remove the JSON file."""
+    model = db.get(ModelBasic, model_id)
+    if not model:
+        return _resp(404, False, "Model not found")
+
+    model_name = model.model_name
+    try:
+        db.delete(model)
+        db.commit()
+    except Exception:
+        db.rollback()
+        logger.exception("Failed to delete model id=%s", model_id)
+        return _resp(500, False, "Failed to delete model")
+
+    # Best-effort removal of the JSON file — do not fail if already gone
+    model_path = os.path.join(MODEL_GENERATION_LOCATION, model_name + MODEL_GENERATION_TYPE)
+    with contextlib.suppress(OSError):
+        os.remove(model_path)
+
+    logger.info("Model '%s' (id=%s) deleted successfully", model_name, model_id)
+    return _resp(200, True, f"Model '{model_name}' deleted successfully")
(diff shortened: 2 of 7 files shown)


NamanmeetSingh:
Incredible work on this, Kartikey. The implementation is exceptionally clean across the stack.

I was specifically looking at the backend code to see how you handled potential file locks (e.g., if the OS throws a `PermissionError` while trying to delete the JSON file, or if it was already deleted). Using `contextlib.suppress(OSError)` is the absolute perfect, Pythonic way to handle that best-effort cleanup without crashing the request pipeline. Brilliant implementation.

The React state management to clear `selectedModel` on deletion and the `MagicMock` unit tests are superb as well.

One edge-case thought for our broader architecture (definitely not a blocker for this PR): If a user deletes a model while it is actively training, the DB cascade will wipe the configs. We might eventually need a mechanism to gracefully kill the training thread/Socket.IO loop if it tries to write metrics to a model that no longer exists in the DB.

Left a review, LGTM!

ivantha (changes requested):
Feature looks good overall. Three things to address:

  1. Tests don't run without TensorFlow installed. The PR says "no live DB or TensorFlow required" but from app.services.deep_learning import delete_model_service pulls in import tensorflow as tf at the module level, so the tests fail with ModuleNotFoundError. You'd need to either extract delete_model_service into its own module or mock the TF import.
  2. Ruff lint failures in tests/test_deep_learning_service.py. import os is unused and imports are unsorted. Running ruff check on the test file returns 2 errors (F401, I001).
  3. No user feedback on delete failure. In Training.jsx, if deleteModel() fails the dialog just closes silently. The user has no idea the deletion didn't work. Would be good to show an error message or toast.

  Everything else looks solid. The cascade cleanup, confirmation dialog, and state management are all handled
  correctly.

ivantha (approved):

PR #172, opened 2026-02-26
fix: wire batch_size parameter through full training stack (#127)

### LINKED ISSUE #127: Wire batch_size Parameter Through to Training
**Difficulty:** Medium
**Area:** full stack

## Description

The Training page in `Training.jsx` includes a `batch_size` field in the training configuration form, allowing users to specify how many samples should be processed per gradient update. However, this value is not actually passed through to the backend's `model.fit()` call. The batch size is collected from the user but silently dropped somewhere in the data flow, causing TensorFlow to use its default batch size of 32 regardless of what the user specifies.

This issue requires tracing the full data flow from the frontend form through the API request, backend schema validation, endpoint handler, and finally into the `model.fit()` call in `model_run.py`, wiring the `batch_size` parameter at each stage.

## Acceptance Criteria

- [ ] The batch_size value entered in the training form is included in the API request body
- [ ] The backend training request schema accepts and validates batch_size (integer, positive, optional with default 32)
- [ ] The backend training logic uses the user-specified batch_size when fitting the model
- [ ] Training with batch_size=16 produces different training behavior than batch_size=128 (verifiable by epoch duration and number of batches)
- [ ] Default behavior (batch_size=32) is preserved when no value is specified

## Resources

- `tensormap-frontend/src/containers/Training/Training.jsx` — training configuration form
- `tensormap-frontend/src/services/ModelServices.js` — API service calls
- `tensormap-backend/app/schemas/deep_learning.py` — request/response schemas
- `tensormap-backend/app/routers/deep_learning.py` — training endpoint
- `tensormap-backend/app/services/model_run.py` — `model_run()` function with `model.fit()`
- [tf.keras.Model.fit() documentation](https://www.tensorflow.org/api_docs/python/tf/keras/Model#fit)

kartikeyg0104:
### Summary

Closes #127

The `batch_size` field was collected from the user on the Training page but silently dropped before ever reaching `model.fit()`. TensorFlow always used its default of 32 regardless of what was entered. This PR wires `batch_size` through every layer of the stack so the user's value is actually respected.

### Root Cause

`batch_size` was lost at four independent points:

| Layer | Problem |
|-------|---------|
| `Training.jsx` `handleSaveConfig` | `batch_size` existed in state but was never included in the payload sent to the API |
| `TrainingConfigRequest` schema | No `batch_size` field — Pydantic would reject or ignore any value sent |
| `update_training_config_service` | Never assigned `config["batch_size"]` to `model.batch_size` |
| `ModelBasic` DB model | No `batch_size` column — nothing to persist or read back in `model_run.py` |

#### Backend

| File | Change |
|------|--------|
| `app/models/ml.py` | Added `batch_size: int \| None` nullable column to `ModelBasic` |
| `migrations/versions/c3a1d9e02f45_...py` | Alembic migration to add the column (`down_revision = "f1a2b3c4d5e6"`, the current single head on main) |
| `app/schemas/deep_learning.py` | Added `batch_size: int = Field(default=32, gt=0)` to `TrainingConfigRequest` |
| `app/services/deep_learning.py` | `update_training_config_service` now persists `model.batch_size = config.get("batch_size", 32)` |
| `app/services/model_run.py` | Reads `model_configs.batch_size` (falls back to 32 for legacy rows) and passes it to `model.fit()` in the tabular path |

#### Frontend

| File | Change |
|------|--------|
| `Training.jsx` | `handleSaveConfig` now includes `batch_size: trainingConfig.batch_size ? Number(trainingConfig.batch_size) : 32` in the PATCH request payload |

### Backward Compatibility

- Existing model rows with `batch_size = NULL` fall back to **32** at runtime — identical behaviour to before
- API clients that omit `batch_size` receive the default of 32 via Pydantic's field default
- Migration column is nullable so `alembic upgrade head` is non-breaking for existing data
- No breaking changes to any existing endpoint or schema

### Testing

- Manually verified the field flows end-to-end: form → PATCH request → DB → `model.fit()`
- Setting `batch_size=16` vs `batch_size=128` on the same dataset produces visibly different epoch durations and steps-per-epoch counts in the Socket.IO progress output

### DIFF
--- a/tensormap-backend/app/models/ml.py
+++ b/tensormap-backend/app/models/ml.py
@@ -29,6 +29,7 @@ class ModelBasic(SQLModel, table=True):
     optimizer: str | None = Field(default=None, max_length=50, nullable=True)
     metric: str | None = Field(default=None, max_length=50, nullable=True)
     epochs: int | None = Field(default=None, nullable=True)
+    batch_size: int | None = Field(default=None, nullable=True)
     loss: str | None = Field(default=None, max_length=50, nullable=True)
     created_on: datetime | None = Field(default=None, sa_column=Column(DateTime, server_default=func.now()))
     updated_on: datetime | None = Field(
--- a/tensormap-backend/app/schemas/deep_learning.py
+++ b/tensormap-backend/app/schemas/deep_learning.py
@@ -97,4 +97,5 @@ class TrainingConfigRequest(BaseModel):
     optimizer: str = Field(min_length=1)
     metric: str = Field(min_length=1)
     epochs: int = Field(gt=0)
+    batch_size: int = Field(default=32, gt=0)
     project_id: uuid_pkg.UUID | None = None
--- a/tensormap-backend/app/services/deep_learning.py
+++ b/tensormap-backend/app/services/deep_learning.py
@@ -241,6 +241,7 @@ def update_training_config_service(
     model.optimizer = config["optimizer"]
     model.metric = config["metric"]
     model.epochs = config["epochs"]
+    model.batch_size = config.get("batch_size", 32)
     model.loss = loss
 
     try:
--- a/tensormap-backend/app/services/model_run.py
+++ b/tensormap-backend/app/services/model_run.py
@@ -171,6 +171,8 @@ def _run(model_name: str, db: Session) -> None:
         x_testing = X[split_index:]
         y_testing = y[split_index:]
 
+        batch_size = model_configs.batch_size if model_configs.batch_size is not None else 32
+
     with open(_helper_generate_json_model_file_location(model_name=model_name)) as f:
         json_string = f.read()
     model = tf.keras.models.model_from_json(json_string, custom_objects=None)
@@ -199,6 +201,7 @@ def _run(model_name: str, db: Session) -> None:
             x_training,
             y_training,
             epochs=model_configs.epochs,
+            batch_size=batch_size,
             callbacks=[CustomProgressBar()],
             verbose=0,
         )
--- a/tensormap-backend/migrations/versions/c3a1d9e02f45_add_batch_size_to_model_basic.py
+++ b/tensormap-backend/migrations/versions/c3a1d9e02f45_add_batch_size_to_model_basic.py
@@ -1,0 +1,27 @@
+"""add_batch_size_to_model_basic
+
+Revision ID: c3a1d9e02f45
+Revises: f1a2b3c4d5e6
+Create Date: 2026-02-27 00:00:00.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "c3a1d9e02f45"
+down_revision = "f1a2b3c4d5e6"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "model_basic",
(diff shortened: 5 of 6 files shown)


NamanmeetSingh:
Really nice work wiring this up end-to-end, Kartikey. The Alembic migration and Pydantic default fallbacks are spot on for backward compatibility.

I was reviewing the `model_run.py` changes and noticed one critical edge case regarding the two different data pipelines. It looks like `model.batch_size` is successfully being passed to `model.fit()` for the tabular/CSV dataset path.

However, for `ProblemType.IMAGE_CLASSIFICATION`, TensorMap uses `tf.keras.utils.image_dataset_from_directory`. Because `tf.data.Dataset` objects are pre-batched, Keras ignores the `batch_size` argument in `model.fit()` if passed. Currently, the image pipeline pulls its batch size directly from the `ImageProperties` DB table during dataset creation.

If a user sets a new `batch_size` of `64` in the Training UI for an image model, this implementation will save it to `ModelConfigs` but the image generator will still yield batches based on the original `ImageProperties`.

We probably need to update the `image_dataset_from_directory` call in the:

```python
if model_configs.model_type == ProblemType.IMAGE_CLASSIFICATION:
````

block to use this new `model_configs.batch_size` (perhaps falling back to `image_properties.batch_size` if `null`) so both data pipelines respect the user's UI input.

Great foundation here though.

ivantha (changes requested):
The batch_size wiring (the actual fix for #127) looks correct -- schema, migration, service layer, and model.fit() are all properly connected with sensible fallbacks. Nice work tracing the full data flow.

However, **the build is broken**. `Training.jsx` imports `deleteModel` from `ModelServices.js` but that function doesn't exist in the file. Running `vite build` confirms the failure:

  > "deleteModel" is not exported by "src/services/ModelServices.js"

There's also incomplete scaffolding for a delete-model feature: `delete_model_service` exists in the backend but has no route, there's no `BACKEND_DELETE_MODEL` URL constant, `Trash2` is imported but never rendered, and `handleDeleteClick` is defined but never called from JSX.

To fix this, please either:
  1. Remove all the delete-model code from this PR (keep it focused on #127), or
  2. Complete the delete feature in a separate PR and rebase

The model list response format change (strings to objects) is also a separate concern that should go with the delete feature PR, not here.

Once the out-of-scope code is removed, the batch_size fix itself looks good to merge.

ivantha (changes requested):
The migration c3a1d9e02f45 has down_revision = "2e66fba94864", which creates a branch in the Alembic chain. There are already two migrations after that revision (f4a8c2e91b37 and f1a2b3c4d5e6), so this gives us two heads and alembic upgrade head will fail.

Fix: change down_revision in c3a1d9e02f45 to "f1a2b3c4d5e6" (the current single head on main).

Everything else looks good.

kartikeyg0104:
> The migration c3a1d9e02f45 has down_revision = "2e66fba94864", which creates a branch in the Alembic chain. There are already two migrations after that revision (f4a8c2e91b37 and f1a2b3c4d5e6), so this gives us two heads and alembic upgrade head will fail.
> 
> Fix: change down_revision in c3a1d9e02f45 to "f1a2b3c4d5e6" (the current single head on main).
> 
> Everything else looks good.

Good catch just fixed it.

ivantha (approved):

PR #175, opened 2026-02-27
feat: Implement model deletion functionality and update model listing

### LINKED ISSUE #133: Model Deletion with Cascade Cleanup

anudeeps0306:
## Description

I have implemented the model deletion feature for both backend and frontend.

Backend: Created the delete_model_service to accept an ID, query the model configs, cascade deletes across the DB schema for configs, remove any generated .json files, and successfully hooked it locally to the router /model/{id} path via HTTP DELETE.
Backend Model Format: Modified get_available_model_list format to dynamically return an array of objects [{"id": "X", "name": "Y"}] instead of just an array of ["Y"].
Frontend Model Services: Developed the API function deleteModel resolving with promises to the Backend DELETE route.
Frontend UI: Appended the requested model Delete visual feature with a React Trash2 Icon component directly adjacent to the model dropdown inside Training.jsx. It properly spawns a clean shadcn/ui confirmation dialog verifying intent to completely wipe the model.
Verification: Fixed the volume paths on docker-compose to hot-reload, successfully wrote to the database to create a sample test model, extracted the dynamically issued sequence-generated ID, successfully triggered the new deletion endpoint via our test script, and verified HTTP 200 response alongside confirming the empty array return list post-deletion.

Fixes #133

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Describe the tests you ran to verify your changes.

- [ ] Existing tests pass
- [ ] New tests added
- [x] Manual testing

## Checklist

- [ ] Tests pass locally

### DIFF
--- a/tensormap-backend/app/services/deep_learning.py
+++ b/tensormap-backend/app/services/deep_learning.py
@@ -360,7 +360,39 @@ def get_available_model_list(
     total = db.exec(select(func.count()).select_from(base_filter.subquery())).one()
 
     models = db.exec(base_filter.offset(offset).limit(limit)).all()
-    data = [m.model_name for m in models]
+    data = [{"id": str(m.id), "name": m.model_name} for m in models]
     body = {"success": True, "message": "Model list generated successfully.", "data": data}
     body["pagination"] = {"total": total, "offset": offset, "limit": limit}
     return body, 200
+
+def delete_model_service(db: Session, model_id: int, project_id: uuid_pkg.UUID | None = None) -> tuple:
+    """Delete a model and its associated configurations."""
+    stmt = select(ModelBasic).where(ModelBasic.id == model_id)
+    if project_id is not None:
+        stmt = stmt.where(ModelBasic.project_id == project_id)
+    model = db.exec(stmt).first()
+    
+    if not model:
+        return _resp(404, False, "Model not found")
+
+    model_name = model.model_name
+
+    try:
+        db.delete(model)
+        db.commit()
+    except Exception as e:
+        db.rollback()
+        logger.exception("Error deleting model")
+        return _resp(500, False, f"Failed to delete model from database: {str(e)}")
+
+    # Try to clean up generated model file if it exists
+    model_path = os.path.join(MODEL_GENERATION_LOCATION, model_name + MODEL_GENERATION_TYPE)
+    if os.path.exists(model_path):
+        try:
+            os.remove(model_path)
+            logger.info("Deleted model file %s", model_path)
+        except OSError:
+            logger.warning("Failed to delete generated model file %s", model_path)
+
+    logger.info("Model '%s' (ID: %d) deleted successfully", model_name, model_id)
+    return _resp(200, True, "Model deleted successfully")
--- a/tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx
+++ b/tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx
@@ -138,14 +138,17 @@ function Canvas() {
     saveModel(data)
       .then((resp) => {
         if (resp.success) {
-          setModelList((prevList) => [
-            ...prevList,
-            {
-              label: modelName + strings.MODEL_EXTENSION,
-              value: modelName,
-              key: prevList.length + 1,
-            },
-          ]);
+          getAllModels(projectId)
+            .then((response) => {
+              const models = response.map((file, index) => ({
+                label: file.name + strings.MODEL_EXTENSION,
+                value: file.name,
+                id: file.id,
+                key: index,
+              }));
+              setModelList(models);
+            })
+            .catch((error) => logger.error("Error refreshing models:", error));
         }
         setFeedbackDialog({
           open: true,
--- a/tensormap-frontend/src/services/ModelServices.js
+++ b/tensormap-frontend/src/services/ModelServices.js
@@ -4,6 +4,27 @@ import * as strings from "../constants/Strings";
 import logger from "../shared/logger";
 
 /**
+ * Deletes a model by its ID.
+ *
+ * @param {string|number} modelId
+ * @param {string} [projectId]
+ * @returns {Promise<{ success: boolean, message: string }>}
+ */
+export const deleteModel = async (modelId, projectId) => {
+  const params = projectId ? { project_id: projectId } : {};
+  return axios
+    .delete(`${urls.BACKEND_GET_MODEL_GRAPH}/${encodeURIComponent(modelId)}`, { params })
+    .then((resp) => resp.data)
+    .catch((err) => {
+      logger.error(err);
+      if (err.response && err.response.data) {
+        return err.response.data;
+      }
+      return { success: false, message: "Unknown error occurred" };
+    });
+};
+
+/**
  * Sends a model graph and training config to the backend for validation.
  *
  * On network error, returns a fallback `{ success: false }` object
--- a/tensormap-backend/app/routers/deep_learning.py
+++ b/tensormap-backend/app/routers/deep_learning.py
@@ -9,6 +9,7 @@ from sqlmodel import Session
 from app.database import get_db
 from app.schemas.deep_learning import ModelNameRequest, ModelSaveRequest, ModelValidateRequest, TrainingConfigRequest
 from app.services.deep_learning import (
+    delete_model_service,
     get_available_model_list,
     get_code_service,
     get_model_graph_service,
@@ -22,6 +23,18 @@ from app.shared.logging_config import get_logger
 logger = get_logger(__name__)
 
 router = APIRouter(tags=["deep-learning"])
+
+
+@router.delete("/model/{model_id}")
+async def delete_model(
+    model_id: int,
+    project_id: uuid_pkg.UUID | None = Query(None),
+    db: Session = Depends(get_db),
+):
+    """Delete a saved model and its configurations."""
+    logger.debug("Deleting model with id=%d", model_id)
+    body, status_code = delete_model_service(db, model_id=model_id, project_id=project_id)
+    return JSONResponse(status_code=status_code, content=body)
 
 
 @router.post("/model/validate")
--- a/tensormap-frontend/src/containers/Training/Training.jsx
+++ b/tensormap-frontend/src/containers/Training/Training.jsx
@@ -22,9 +22,19 @@ import {
   runModel,
   getAllModels,
   updateTrainingConfig,
+  deleteModel,
 } from "../../services/ModelServices";
 import { getAllFiles } from "../../services/FileServices";
 import { models as modelListAtom } from "../../shared/atoms";
+import { Trash2 } from "lucide-react";
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from "@/components/ui/dialog";
 
 const optimizerOptions = [
   { key: "opt_1", label: "Adam", value: "adam" },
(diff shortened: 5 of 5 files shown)


tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx
```diff
@@ -246,20 +246,17 @@ function Canvas() {
...
-          setModelList((prevList) => [
-            ...prevList,
-            {
-              label: modelName + strings.MODEL_EXTENSION,
-              value: modelName,
-              key: prevList.length + 1,
-            },
-          ]);
+          getAllModels(projectId)
+            .then((response) => {
```
rajutkarsh07: draft not cleared after save
The new save success handler refreshes models but removed draft cleanup.
Result: stale draft remains in `localStorage` and “Discard Draft” stays visible after save.

**Fix:** restore cleanup before `getAllModels`:

```js
if (resp.success) {
  try {
    localStorage.removeItem(draftKey);
    setHasDraft(false);
  } catch (e) {
    logger.error("Failed to clear draft on save:", e);
  }
  // then refresh models…
}
```

ivantha (commented):
## ◎ Argus Review

This PR implements model deletion across the backend and frontend stack, including a REST endpoint, service layer, and a confirmation dialog UI. The implementation is functional for the happy path, but carries several significant concerns: sequential integer model IDs with no ownership or authentication check introduce an IDOR vulnerability; unsanitized model names in file-system operations pose a path-traversal risk; and a meaningful regression removes the localStorage draft-clearing logic after a model save. The frontend error handling also silently swallows errors without surfacing feedback to users.

tensormap-frontend/src/services/ModelServices.js
```diff
@@ -3,6 +3,27 @@ import * as urls from "../constants/Urls";
...
+export const deleteModel = async (modelId, projectId) => {
+  const params = projectId ? { project_id: projectId } : {};
+  return axios
+    .delete(`${urls.BACKEND_GET_MODEL_GRAPH}/${encodeURIComponent(modelId)}`, { params })
+    .then((resp) => resp.data)
+    .catch((err) => {
+      logger.error(err);
+      if (err.response && err.response.data) {
+        return err.response.data;
+      }
```
ivantha: **[MEDIUM]** Error Swallowing Makes the Component's .catch() Branch Unreachable

The service's `.catch()` catches all Axios errors and returns a **resolved** promise:
```js
.catch((err) => {
  logger.error(err);
  if (err.response && err.response.data) {
    return err.response.data;   // resolved, not rejected
  }
  return { success: false, message: "Unknown error occurred" };
});
```
From the caller's perspective (`Training.jsx`), the promise **never rejects**. The `.catch((error) => { … })` in the component is dead code for any HTTP-level error. All error paths flow through `.then((resp) => { if (resp.success) … })`, which is easy to miss and means the component's catch handler provides no actual safety net.

**Suggestion:**
```
Decide on one consistent error-handling contract. The simplest approach that enables the component's `.catch()` to work is to re-throw (or propagate a rejected promise) on unrecoverable errors, and only return the response object for structured API errors:

```js
export const deleteModel = async (modelId, projectId) => {
  const params = projectId ? { project_id: projectId } : {};
  const resp = await axios.delete(
    `${urls.BACKEND_MODEL_BASE}/${encodeURIComponent(modelId)}`,
    { params }
  );
  return resp.data;  // let Axios throw on non-2xx; component .catch() will fire
};
```
Alternatively, keep the swallowing pattern but **remove** the dead `.catch()` in the component and handle all outcomes in `.then()`.
```

File: tensormap-backend/app/services/deep_learning.py [hunk 3]
```diff
@@ -360,7 +360,39 @@ def get_available_model_list(
     total = db.exec(select(func.count()).select_from(base_filter.subquery())).one()
 
     models = db.exec(base_filter.offset(offset).limit(limit)).all()
-    data = [m.model_name for m in models]
+    data = [{"id": str(m.id), "name": m.model_name} for m in models]
```
ivantha: **[LOW]** Model ID Returned as String but Backend Route Expects Integer — Type Inconsistency

The listing endpoint now returns each model ID as a Python string:
```python
data = [{"id": str(m.id), "name": m.model_name} for m in models]
```
The frontend stores this string and passes it to `deleteModel(modelToDelete.id, projectId)`, which embeds it in the URL. While this happens to work because HTTP URLs are strings, it creates a confusing contract: the DB column and the route parameter are integers (`model_id: int`), but the listing API advertises them as strings.

This may also cause comparison bugs if any code ever uses `===` equality between the stored string ID and a numeric ID.

**Suggestion:**
```
Return the numeric ID directly (no `str()` cast) and document the field type in an API schema/serialiser:

```python
data = [{"id": m.id, "name": m.model_name} for m in models]
```
On the frontend, if TypeScript is ever adopted or prop-types are added, the type will be clearly `number`.
```

File: tensormap-backend/app/services/deep_learning.py [hunk 4]
```diff
@@ -360,7 +360,39 @@ def get_available_model_list(
...
+
+    model_name = model.model_name
+
+    try:
+        db.delete(model)
+        db.commit()
+    except Exception as e:
+        db.rollback()
+        logger.exception("Error deleting model")
+        return _resp(500, False, f"Failed to delete model from database: {str(e)}")
```
ivantha: **[LOW]** Overly Broad Exception Catch Hides Programming Errors

```python
except Exception as e:
    db.rollback()
    logger.exception("Error deleting model")
    return _resp(500, False, f"Failed to delete model from database: {str(e)}")
```
`Exception` is the base class for almost everything in Python, including `AttributeError`, `TypeError`, and other bugs that would normally surface as 500s during development. Catching `Exception` broadly masks these programming mistakes and converts them into misleading "database error" messages.

**Suggestion:**
```
Catch only database-specific errors:

```python
from sqlalchemy.exc import SQLAlchemyError

try:
    db.delete(model)
    db.commit()
except SQLAlchemyError:
    db.rollback()
    logger.exception("Database error deleting model id=%d", model_id)
    return _resp(500, False, "Failed to delete model. Please try again.")
```
```

File: tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx [hunk 5]
```diff
@@ -246,20 +246,17 @@ function Canvas() {
...
+                value: file.name,
+                id: file.id,
+                key: index,
+              }));
+              setModelList(models);
+            })
+            .catch((error) => logger.error("Error refreshing models:", error));
         }
         setFeedbackDialog({
           open: true,
```
ivantha: **[LOW]** Nested Promise Anti-pattern — Should Use async/await or Flat Chain

After `saveModel(data).then(...)`, an inner `getAllModels(...).then(...).catch(...)` is nested within the outer `.then()` callback. This "pyramid of doom" pattern makes error propagation unclear and is hard to reason about. In particular, the outer `.then()`'s error handling does not cover the inner `getAllModels` failure path beyond a bare `logger.error`.

**Suggestion:**
```
Refactor to async/await for clarity:

```js
saveModel(data)
  .then(async (resp) => {
    if (resp.success) {
      try {
        localStorage.removeItem(draftKey);
        setHasDraft(false);
      } catch (e) {
        logger.error("Failed to clear draft on save:", e);
      }
      try {
        const response = await getAllModels(projectId);
        const models = response.map((file) => ({
          label: file.name + strings.MODEL_EXTENSION,
          value: file.name,
          id: file.id,
          key: file.id,
        }));
        setModelList(models);
      } catch (error) {
        logger.error("Error refreshing models:", error);
      }
    }
    // ...
  });
```
```

File: tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx [hunk 6]
```diff
@@ -246,20 +246,17 @@ function Canvas() {
...
-            },
-          ]);
+          getAllModels(projectId)
+            .then((response) => {
+              const models = response.map((file, index) => ({
+                label: file.name + strings.MODEL_EXTENSION,
+                value: file.name,
+                id: file.id,
+                key: index,
+              }));
```
ivantha: **[LOW]** Array Index Used as React key — Unstable on List Mutations

Both Canvas.jsx and Training.jsx map models to list items using `key: index` (or `key: index` derived from `.map((file, index) => ...)`). React uses `key` to track element identity across renders. Using the array index means that after a deletion, all items after the removed element get new keys, forcing unnecessary DOM reconciliation and potentially causing transient rendering bugs with controlled components (e.g., selected dropdown values).

This is especially problematic here because the list is mutated by deletion.

**Suggestion:**
```
Use the stable model `id` field as the key:

```js
const models = response.map((file) => ({
  label: file.name + strings.MODEL_EXTENSION,
  value: file.name,
  id: file.id,
  key: file.id,   // ← stable, unique
}));
```
```

File: tensormap-frontend/src/services/ModelServices.js [hunk 7]
```diff
@@ -3,6 +3,27 @@ import * as urls from "../constants/Urls";
 import * as strings from "../constants/Strings";
 import logger from "../shared/logger";
 
+/**
+ * Deletes a model by its ID.
+ *
+ * @param {string|number} modelId
```
ivantha: **[INFO]** async Function Declared but await Is Never Used

```js
export const deleteModel = async (modelId, projectId) => {
  const params = projectId ? { project_id: projectId } : {};
  return axios
    .delete(...)  // promise returned, not awaited
```
Declaring a function `async` without any `await` expressions is misleading — it implicitly wraps the return value in a `Promise.resolve()`, but so does a regular function that already returns a Promise. The `async` keyword adds no value here and may confuse readers into thinking there is an `await` somewhere.

**Suggestion:**
```
Either remove `async` (since Axios already returns a Promise):
```js
export const deleteModel = (modelId, projectId) => {
  ...
};
```
Or convert fully to async/await style (see the error-swallowing finding above for a complete rewrite).
```

ivantha:
CI is failing on this PR: both **Frontend Lint** and **Backend Lint** checks are reporting errors. Please run `npm run lint` in the frontend and `uv run ruff check . && uv run ruff format --check .` in the backend, fix the reported issues, and push an update.

ivantha:
Thank you for implementing model deletion! This feature was already shipped in merged PR #170 (feat: add model deletion with cascade cleanup). This PR has accumulated change requests and conflicts with main. We appreciate the contribution and encourage you to look at other open issues.

PR #180, opened 2026-02-28
feat(frontend): implement dynamic SSOT layer UI and generic nodeFeature/ssot frontend nodes

### LINKED ISSUE #177: [Architecture Proposal] Phase 1 Foundation: Dynamic Data-Driven Layer Registry (Single Source of Truth)
## Overview

As outlined in the GSoC 2026 Phase 1 Expected Results, TensorMap needs to transition from a hardcoded layer system (requiring 6+ file edits) to a data-driven layer registry. 

To achieve this and easily scale to 15+ layer types (CNNs, RNNs, Regularization), I propose a **Single Source of Truth (SSOT)** architecture. By defining layers in a central JSON registry served by the backend, we can decouple the frontend UI from the backend generation logic.

## Current Bottleneck (The 6-File Problem)

Currently, adding a layer like `Dropout` requires:

1. Hardcoding the button in `Sidebar.jsx`
2. Creating a dedicated `DropoutNode.jsx` UI component
3. Registering the node in `Canvas.jsx`
4. Hardcoding Pydantic schemas in `deep_learning.py`
5. Updating the code generation engine
6. Updating the Keras compilation mapping

## Proposed Architecture: The Unified Registry

I propose creating a `layer_registry.json` on the backend and serving it via a new endpoint (`GET /api/v1/layers`). Both the FastAPI backend and React frontend will use this single contract.

### 1. The JSON Contract (Example)

```json
{
  "layers": {
    "dense": {
      "display_name": "Dense",
      "category": "core",
      "keras_mapping": "tf.keras.layers.Dense",
      "params": {
        "units": { "type": "number", "default": 32, "min": 1, "required": true },
        "activation": { "type": "select", "options": ["relu", "sigmoid", "softmax", "linear"], "default": "relu" }
      }
    },
    "dropout": {
      "display_name": "Dropout",
      "category": "regularization",
      "keras_mapping": "tf.keras.layers.Dropout",
      "params": {
        "rate": { "type": "number", "default": 0.5, "min": 0.0, "max": 1.0, "required": true }
      }
    }
  }
}
````

### 2. Frontend Implementation (Dynamic Rendering)

**Dynamic Sidebar:**
`Sidebar.jsx` fetches the registry on load and maps through layers to generate the draggable buttons, grouping them by category.

**Universal Node Component:**
Instead of a separate `.jsx` file for every layer, we build a single `GenericLayerNode.jsx`. When dropped on the canvas, it looks up its type in the registry and dynamically maps the `params` object to UI inputs (rendering a `<Select>` for `activation` and a number `<Input>` for `units`).

### 3. Backend Implementation (Dynamic Validation)

**Pydantic Dynamic Models:**
Instead of hardcoded models, we use `create_model` in Pydantic to dynamically generate validation schemas based on the JSON registry at server startup.

**Streamlined Code Gen:**
The generation engine loops through the ReactFlow payload, looks up the `keras_mapping` string, and dynamically unpacks the frontend payload into kwargs:

```python
f"{layer.keras_mapping}(**{layer.params})"
```
## Action Plan & Phased Rollout

If the maintainers are aligned with this architecture, I am ready to lead Phase 1 with the following PRs:

* **PR 1:** Define the `layer_registry.json` and build the `GET /api/v1/layers` endpoint.
* **PR 2:** Build the frontend `GenericLayerNode.jsx` to dynamically consume the API and replace the existing hardcoded Dense/Conv2D nodes.
* **PR 3:** Update backend code generation to use the registry dictionary.

@ivantha I'd love your thoughts on this SSOT approach. Does this align with your vision for the Phase 1 Foundation?

NamanmeetSingh:
## Description

As step 2 of the Phase 1 Foundation architecture, this PR migrates the frontend to dynamically consume the backend Layer Registry API, eliminating hardcoded layer definitions from the React UI.

**Issues Addressed:**
* **Scalability Bottleneck:** Previously, adding a single new deep learning layer required manually creating new React components, wiring them into the canvas schema, and hardcoding their parameter forms.
* **Maintenance Debt:** The UI parameters and backend generation logic were decoupled, leading to potential mismatches. This PR forces the frontend to build its UI strictly from the backend's SSOT JSON contract.

**Specific Changes Made:**
* **`Urls.js` & `ModelServices.js`**: Added the fetch logic to hit the new `GET /api/v1/layers` endpoint.
* **`Sidebar.jsx`**: Replaced the hardcoded drag-and-drop layer list with a dynamic map that generates buttons based on the API's registry categories.
* **`Canvas.jsx`**: 
  * Replaced the hardcoded `nodeTypes` dictionary with a single `genericLayer`.
  * Updated the `onDrop` handler to automatically parse default parameters from the backend schema.
  * **Fix:** Resolved the React Flow `project()` deprecation warning by migrating to `screenToFlowPosition`.
* **`NodePropertiesPanel.jsx`**: Completely rewrote the 200+ line hardcoded property panel. It now dynamically loops through the selected node's schema, generating integer, float, and string inputs, as well as selection dropdowns, entirely on the fly.
* **Custom Nodes**: Deleted `DenseNode.jsx`, `ConvNode.jsx`, `FlattenNode.jsx`, and `InputNode.jsx` and replaced them all with a single, universal `GenericLayerNode.jsx`.

Fixes #177 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Manually tested by running both the Uvicorn backend and React frontend.
- [x] Verified the Sidebar dynamically renders available layers from the API.
- [x] Dragged layers onto the canvas and verified `GenericLayerNode` renders with correct default parameters.
- [x] Selected nodes and verified `NodePropertiesPanel` dynamically generates inputs and dropdowns based on the layer schema.
- [x] Verified no `customnode` console errors remain on a fresh canvas.

Screenshot 2026-02-28 211651

## Checklist

- [ ] I have added/updated documentation as needed
- [x] Tests pass locally

## Breaking Changes
- Existing saved canvas drafts and models will need migration: node `data.registry` must be re-hydrated from the new registry API on load.
- Parameter key names have changed for Input (dim-1→dim_1) and Conv2D (filter→filters, kernelX→kernel_size_rows/cols). Any persisted model JSON must be migrated.
- Conv2D stride and padding parameters are temporarily absent from the new registry.


NamanmeetSingh:
Thanks for the incredibly thorough review, @rajutkarsh07. You caught some great edge cases. I've pushed the fixes for everything mentioned. 

Here is the breakdown of the updates:

- [x] **Reverted `docker-compose.yml` and `alembic.ini`**: Restored these files to their exact `main` state to prevent breaking the shared Docker environment and avoid leaking local credentials.
- [x] **Restored backward compatibility**: Added mappings for `custominput`, `customdense`, `customconv`, and `customflatten` to `nodeTypes` so existing saved models in the DB will seamlessly render using the new `GenericLayerNode`.
- [x] **Fixed Input layer connections**: Added conditional rendering to hide the `_in` target handle specifically for Input layers, preventing invalid backward connections.
- [x] **Fixed `getLayerRegistry` API call**: Removed the hardcoded localhost `fetch` and updated it to correctly use the configured `axios` client with the `BACKEND_GET_LAYERS` constant.
- [x] **Added Input registry params**: Updated `layer_registry.json` to include `dim_1`, `dim_2`, and `dim_3` so the UI correctly generates the tensor shape configuration fields.
- [x] **Fixed `updateParam` type check**: Updated the UI properties panel to correctly cast `"number"` string types to actual numbers, matching the JSON registry contract.
- [x] **Removed duplicate constants**: Cleaned up the redundant `TEMPLATE_ROOT` and `MODEL_GENERATION_LOCATION` definitions at the bottom of `constants.py`.

Everything should be completely synced up and Docker-safe now. Ready for another look.

ivantha (commented):
## ◎ Argus Review

This PR delivers a well-conceived SSOT architecture that successfully eliminates hardcoded layer definitions from the frontend — the overall direction is excellent and the `GenericLayerNode`/dynamic panel approach is a major maintainability win. However, the refactor introduces several high-severity functional bugs that are visible immediately on the happy path: `formatLabel` doesn't handle underscore-separated registry keys (so labels like `dim_1` render literally), all numeric inputs render as HTML `type="text"` due to a type-name mismatch (`"number"` vs `"int"`/`"float"`), the JSON schema's `min` constraints are never applied to inputs, and the backend endpoint uses blocking synchronous file I/O inside an `async` function. These issues should be resolved before merge, along with the `jsconfig.json` `baseUrl` removal which risks breaking IDE path resolution for the entire frontend.

tensormap-frontend/src/services/ModelServices.js
```diff
@@ -2,6 +2,7 @@ import axios from "../shared/Axios";
 import * as urls from "../constants/Urls";
 import * as strings from "../constants/Strings";
 import logger from "../shared/logger";
+import { BACKEND_GET_LAYERS } from "../constants/Urls";
```
ivantha: **[MEDIUM]** Duplicate import of `BACKEND_GET_LAYERS` from the same module

`ModelServices.js` already imports everything from `Urls.js` via `import * as urls from "../constants/Urls"`. The new code then adds a redundant named import:
```javascript
import * as urls from "../constants/Urls";
// ...
import { BACKEND_GET_LAYERS } from "../constants/Urls"; // ← redundant
```
This is redundant and inconsistent with how other URL constants are used throughout the file (where they would be accessed as `urls.BACKEND_...`).

**Suggestion:**
```
Remove the explicit named import and use the namespace import:
```javascript
// Remove: import { BACKEND_GET_LAYERS } from "../constants/Urls";

export const getLayerRegistry = async () => {
  try {
    const resp = await axios.get(urls.BACKEND_GET_LAYERS);
    return resp.data?.data?.layers ?? {};
  } catch (err) {
    logger.error("Failed to fetch layer registry:", err);
    return {};
  }
};
```
```

File: tensormap-frontend/jsconfig.json [hunk 2]
```diff
@@ -1,6 +1,5 @@
 {
   "compilerOptions": {
-    "baseUrl": ".",
     "paths": {
       "@/*": ["./src/*"]
     }
```
ivantha: **[HIGH]** Removing `baseUrl` may break `@/*` path alias resolution

The `jsconfig.json` previously had `"baseUrl": "."` alongside `paths`. According to the TypeScript/JavaScript specification (which `jsconfig.json` follows), `paths` entries are resolved **relative to `baseUrl`**. Removing `baseUrl` makes the `paths` configuration technically non-conformant and can break:

- IDE autocompletion and go-to-definition for `@/...` imports
- Some bundler configurations (e.g., certain Webpack/Vite setups that read `jsconfig.json` for path resolution)
- Jest module name mapping if configured to follow `jsconfig.json`

While Vite itself doesn't use `jsconfig.json` for runtime resolution (it uses its own `resolve.alias` config), the `jsconfig.json` is used by editors like VS Code for intelligence. Removing `baseUrl` will likely cause VS Code to report unresolved `@/...` imports as errors.

**Suggestion:**
```
Restore `baseUrl`:
```json
{
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "@/*": ["./src/*"]
    }
  }
}
```
If `baseUrl` was intentionally removed (e.g., because the project migrated entirely to Vite path aliases configured in `vite.config.js`), document this decision. Otherwise, restore it.
```

File: tensormap-frontend/src/components/DragAndDropCanvas/CustomNodes/GenericLayerNode.jsx [hunk 3]
```diff
@@ -0,0 +1,54 @@
...
+};
+
+function GenericLayerNode({ data, id }) {
+  const { params, registry } = data;
+  
+  const displayName = registry?.display_name || data?.label || "Unknown Layer";
+  const categoryColor = getCategoryColor(registry?.category);
+
+  // Determine if this is an input layer (case-insensitive for safety)
+  const isInput = displayName.toLowerCase().includes("input");
```
ivantha: **[HIGH]** Fragile `isInput` detection via display name string matching

The decision of whether to render the incoming (target) handle is based on whether the word "input" appears in `displayName`:

```javascript
const isInput = displayName.toLowerCase().includes("input");
```

This is fragile for several reasons:
- A future layer named `"Dropout"` → false positive if someone names it `"Input Dropout"`
- A layer named `"InputNormalization"` would incorrectly have no target handle
- An `InputLayer` variant with a different `display_name` (e.g., `"Data Source"`) would incorrectly render with a target handle
- The logic ties a structural graph property ("has incoming connections") to a UI display string, creating a hidden contract

The registry is the Single Source of Truth — this property belongs there.

**Suggestion:**
```
Add an explicit `is_source_only` or `connection_type` field to the registry JSON:
```json
"input": {
  "display_name": "Input",
  "category": "core",
  "is_source_only": true,
  "params": { ... }
}
```
Then use it in the component:
```javascript
const isInput = registry?.is_source_only === true;
```
This makes the topology contract explicit and independent of display naming.
```

File: tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx [hunk 4]
```diff
@@ -29,10 +27,11 @@ import { models as allModels } from "../../shared/atoms";
...
 const nodeTypes = {
-  custominput: InputNode,
-  customdense: DenseNode,
-  customflatten: FlattenNode,
-  customconv: ConvNode,
+  genericLayer: GenericLayerNode,
+  custominput: GenericLayerNode,
+  customdense: GenericLayerNode,
+  customconv: GenericLayerNode,
+  customflatten: GenericLayerNode,
```
ivantha: **[MEDIUM]** Legacy node types in `nodeTypes` map will silently break when old saved nodes are loaded

The `nodeTypes` map retains legacy keys to handle old saved nodes:
```javascript
const nodeTypes = {
  genericLayer: GenericLayerNode,
  custominput: GenericLayerNode,
  customdense: GenericLayerNode,
  customconv: GenericLayerNode,
  customflatten: GenericLayerNode,
};
```
`GenericLayerNode` now expects `data.registry` to be populated (for `display_name`, `category`, `params` schema). Old nodes stored in the database or `localStorage` draft were created without `data.registry`. When loaded, `registry` will be `undefined`, causing:
1. The node title to fall back to `data.label` (e.g., `"custominput node"`) — ugly but functional
2. `NodePropertiesPanel` will render `"Layer Properties"` as title and show `"No configurable parameters"` because `registry.params` is empty — old params are rendered as nothing
3. The `isInput` heuristic may fail if old `data.label` is something like `"custominput node"`

There is no migration path or user warning for this data format change.

**Suggestion:**
```
Add a migration shim in `Canvas.jsx` when loading nodes (e.g., from draft or API). When an old-format node is detected (has type `custominput` etc. but no `data.registry`), reconstruct `registry` from the `layerRegistry` state:

```javascript
// In the draft loading / model loading logic:
const migratedNodes = savedNodes.map(node => {
  if (!node.data.registry && layerRegistry) {
    const legacyKeyMap = { custominput: 'input', customdense: 'dense', customconv: 'conv2d', customflatten: 'flatten' };
    const registryKey = legacyKeyMap[node.type];
    if (registryKey && layerRegistry[registryKey]) {
      return { ...node, type: 'genericLayer', data: { ...node.data, registry: layerRegistry[registryKey] } };
    }
  }
  return node;
});
```
Document this as a known breaking change and consider a DB migration if models are persisted server-side.
```

File: tensormap-backend/app/routers/deep_learning.py [hunk 5]
```diff
@@ -99,3 +101,18 @@ async def get_model_list(
...
+    logger.info("Fetching unified layer registry")
+    try:
+        with open(LAYER_REGISTRY_LOCATION, "r") as f:
+            registry = json.load(f)
+        return JSONResponse(status_code=200, content={"success": True, "data": registry})
+    except FileNotFoundError:
+        logger.error("Layer registry not found at %s", LAYER_REGISTRY_LOCATION)
+        return JSONResponse(
+            status_code=500, 
+            content={"success": False, "message": "Unified layer registry missing from server."}
```
ivantha: **[INFO]** Registry is read from disk on every request — consider startup caching

The `layer_registry.json` is a static file that changes only with deployments. Reading it from disk on every `GET /layers` request is unnecessary I/O, even if fast.

For a high-traffic deployment (or for correctness testing in load tests), this creates repeated disk access for data that will never change at runtime.

**Suggestion:**
```
Cache the registry at application startup using a module-level variable or FastAPI's `lifespan` context:
```python
# In app startup or as a module-level singleton
_layer_registry_cache: dict | None = None

def _load_registry() -> dict:
    global _layer_registry_cache
    if _layer_registry_cache is None:
        with open(LAYER_REGISTRY_LOCATION) as f:
            _layer_registry_cache = json.load(f)
    return _layer_registry_cache

@router.get("/layers")
def get_layer_registry():
    return JSONResponse(content={"success": True, "data": _load_registry()})
```
Or use FastAPI's `@lru_cache` via `functools`.
```

File: tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx [hunk 6]
```diff
@@ -52,6 +51,13 @@ function Canvas() {
     detail: "",
   });
   const [contextMenu, setContextMenu] = useState({ nodeId: null, x: 0, y: 0 });
+  const [layerRegistry, setLayerRegistry] = useState({});
+
+  // Fetch the registry on mount
+  useEffect(() => {
```
ivantha: **[INFO]** No error state surfaced to the user when `getLayerRegistry` fails

The `useEffect` in `Canvas.jsx` calls `getLayerRegistry().then(setLayerRegistry)`. The `getLayerRegistry` service already swallows errors and returns `{}`. So when the backend is down or returns an error, the registry silently remains `{}`, the sidebar is blank, and drops are silently rejected.

This creates a confusing UX where the canvas loads but nothing is draggable, with no explanation.

**Suggestion:**
```
Track error state explicitly:
```javascript
const [registryError, setRegistryError] = useState(null);

useEffect(() => {
  getLayerRegistry()
    .then(data => {
      if (Object.keys(data).length === 0) setRegistryError("Could not load layers from server.");
      setLayerRegistry(data);
    });
}, []);
```
Surface the error as a `toast` notification or an alert banner near the canvas header.
```

File: tensormap-frontend/src/components/DragAndDropCanvas/CustomNodes/GenericLayerNode.jsx [hunk 7]
```diff
@@ -0,0 +1,54 @@
...
+GenericLayerNode.propTypes = {
+  data: PropTypes.shape({
+    params: PropTypes.object.isRequired,
+    registry: PropTypes.object,
+    label: PropTypes.string, // Added to silence prop-types warnings on older nodes
+  }).isRequired,
+  id: PropTypes.string.isRequired,
+};
+
+export default GenericLayerNode;
```
ivantha: **[INFO]** Positive: `GenericLayerNode` is a clean, well-structured universal component

The replacement of four specific node components with a single `GenericLayerNode` is a significant improvement in maintainability. The component correctly:
- Uses optional chaining for safe access to `registry` (handles missing data gracefully)
- Provides meaningful fallbacks (`data?.label || "Unknown Layer"`)
- Computes the summary dynamically without hardcoding param keys
- Separates the color mapping concern cleanly via `getCategoryColor`
- Correctly keeps the source handle always present while conditionally rendering the target handle

The PropTypes comment `// Added to silence prop-types warnings on older nodes` demonstrates awareness of backward compatibility concerns, even if the underlying issue needs a proper migration (see related finding).

**Suggestion:**
```
No change needed for this finding — this is a positive observation. The pattern is solid and extensible.
```

tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx
```diff
@@ -289,43 +295,40 @@ function Canvas() {
...
         return;
       }
 
-      const position = reactFlowInstance.project({
-        x: event.clientX - reactFlowBounds.left,
-        y: event.clientY - reactFlowBounds.top,
+      const position = reactFlowInstance.screenToFlowPosition({
+        x: event.clientX,
+        y: event.clientY,
       });
```
ivantha: **[INFO]** Positive: `screenToFlowPosition` migration from deprecated `project()` is correct

The PR correctly migrates from the React Flow deprecated `reactFlowInstance.project()` API to the new `reactFlowInstance.screenToFlowPosition()` API. The old API required manually subtracting the bounding rect:
```javascript
// OLD (deprecated)
const reactFlowBounds = reactFlowWrapper.current.getBoundingClientRect();
const position = reactFlowInstance.project({
  x: event.clientX - reactFlowBounds.left,
  y: event.clientY - reactFlowBounds.top,
});
```
The new API handles the coordinate transform internally:
```javascript
// NEW (correct)
const position = reactFlowInstance.screenToFlowPosition({
  x: event.clientX,
  y: event.clientY,
});
```
This is the correct fix for the `project()` deprecation warning mentioned in the PR description.

**Suggestion:**
```
No change needed. Remove the dead commented line above this as noted in the separate finding.
```

ivantha (commented):
## ◎ Argus Review

This PR successfully delivers a well-architected SSOT refactor that eliminates hardcoded layer definitions from the frontend and replaces four specialized node components with a single generic one driven by a backend JSON registry. The React Flow `project()` deprecation fix and the dynamic property panel are clean implementations. However, the PR has two critical-severity issues that make it not ready to merge: (1) existing saved models and canvas drafts will silently break because loaded nodes lack the new `data.registry` field that `GenericLayerNode` requires, and (2) the Conv2D registry entry is missing stride and padding parameters, which were present in the old UI and are likely expected by the backend model-generation service. These represent real data-breaking regressions that need migration logic and registry corrections before this can safely reach production.

**Suggested Version Bump:** `major`

### ⚠️ Breaking Changes

- **[CRITICAL]** Existing saved nodes lose their registry data, breaking render and edit: **Previous behavior:** Nodes were saved/loaded with specific types like `custominput`, `customdense`, etc., each rendered by a dedicated component that knew its own param schema.

**New behavior:** `GenericLayerNode` expects `data.registry` to be populated at drop-time from the live API response. Nodes persisted in localStorage (drafts) or the database lack `data.registry`, so they render as `"Unknown Layer"` with no configurable parameters.

**Who is affected:** All existing users who have saved models or canvas drafts. On reload, every existing node silently degrades — the canvas renders something, but editing is lost.

**Migration path:** When loading existing nodes, Canvas should detect the absence of `data.registry` and re-hydrate it from the registry by matching `node.type` (e.g., `customdense`) to the corresponding registry key. A migration map like `{ customdense: 'dense', customconv: 'conv2d', custominput: 'input', customflatten: 'flatten' }` and a re-hydration step in the draft-loading logic would fix this.
  > **Suggestion:** In the draft/model loading logic in Canvas.jsx, add a migration step:
```js
const TYPE_MIGRATION_MAP = {
  custominput: 'input',
  customdense: 'dense',
  customflatten: 'flatten',
  customconv: 'conv2d',
};

const migrateNode = (node, registry) => {
  if (node.data.registry) return node; // already migrated
  const registryKey = TYPE_MIGRATION_MAP[node.type];
  return {
    ...node,
    type: 'genericLayer',
    data: {
      ...node.data,
      registry: registry[registryKey] || {},
    },
  };
};
```
Apply this when restoring nodes from the draft or from the API.
- **[CRITICAL]** Conv2D and Input parameter keys renamed — breaks backend model generation: **Previous behavior:** Parameter keys in node data matched what the backend model-generation service expected:
- `custominput`: `dim-1`, `dim-2`, `dim-3`
- `customconv`: `filter`, `kernelX`, `kernelY`, `strideX`, `strideY`, `padding`, `activation`

**New behavior:** The registry introduces entirely different keys:
- `input`: `dim_1`, `dim_2`, `dim_3` (hyphens → underscores)
- `conv2d`: `filters`, `kernel_size_rows`, `kernel_size_cols`, `activation` (renamed, and **stride/padding are entirely absent**)

**Who is affected:** The backend service that reads `params` from the saved model JSON to generate Keras code. If that service still uses `param['filter']`, `param['kernelX']`, etc., model generation will silently produce incorrect or broken Keras code for any Conv2D layer created after this PR.

**Migration path:** Either update the backend generation code to use the new registry keys, or ensure the registry keys exactly match what the backend already expects. The stride and padding parameters must be added back to the Conv2D registry entry.
  > **Suggestion:** Add missing parameters to `layer_registry.json`:
```json
"conv2d": {
  "params": {
    "filters": { "type": "number", "default": 32, "min": 1, "required": true },
    "kernel_size_rows": { "type": "number", "default": 3, "min": 1, "required": true },
    "kernel_size_cols": { "type": "number", "default": 3, "min": 1, "required": true },
    "stride_rows": { "type": "number", "default": 1, "min": 1, "required": false },
    "stride_cols": { "type": "number", "default": 1, "min": 1, "required": false },
    "padding": { "type": "select", "options": ["valid", "same"], "default": "valid", "required": false },
    "activation": { "type": "select", "options": ["relu", "sigmoid", "softmax", "linear"], "default": "relu", "required": false }
  }
}
```
Then audit the backend Keras-generation service to ensure it maps these new keys correctly.


### PR Description Issues

- **[INFO]** "Breaking change" checkbox left unchecked despite backward-incompatible changes: The PR description has `- [ ] Breaking change` unchecked. However, this PR introduces multiple backward-incompatible changes:
1. Existing saved nodes lose their `registry` data, breaking their render and edit
2. Parameter key names have changed (e.g., `dim-1` → `dim_1`, `filter` → `filters`, `kernelX` → `kernel_size_rows`)
3. Conv2D stride and padding parameters are removed from the new schema

This misinforms reviewers and automated release tooling.
  > **Suggestion:** Update the type-of-change section:
```markdown
## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Breaking change  ← check this
- [ ] Documentation update

## Breaking Changes
- Existing saved canvas drafts and models will need migration: node `data.registry` must be re-hydrated from the new registry API on load.
- Parameter key names have changed for Input (dim-1→dim_1) and Conv2D (filter→filters, kernelX→kernel_size_rows/cols). Any persisted model JSON must be migrated.
- Conv2D stride and padding parameters are temporarily absent from the new registry.
```
- **[INFO]** Testing section covers only happy-path manual flows: The "How Has This Been Tested?" section only describes successful happy-path scenarios. It doesn't mention testing:
- What happens when the `/api/v1/layers` endpoint is down
- Loading an existing saved model/draft created before this PR
- The Conv2D layer (which is missing stride/padding, so drag-and-drop would succeed but model generation would fail)
- Any automated test execution
  > **Suggestion:** Expand the testing section:
```markdown
## How Has This Been Tested?
- [x] Verified the Sidebar dynamically renders available layers from the API.
- [x] Dragged layers onto the canvas and verified `GenericLayerNode` renders with correct default parameters.
- [x] Selected nodes and verified `NodePropertiesPanel` dynamically generates inputs and dropdowns.
- [x] Verified no `customnode` console errors remain on a fresh canvas.
- [ ] Tested loading a canvas draft saved before this PR (backward compatibility) — **NOT TESTED**
- [ ] Tested behavior when backend `/layers` endpoint returns 500 — **NOT TESTED**
- [ ] Ran existing test suite: `npm test` / `pytest` — **NO NEW TESTS ADDED**
```
- **[INFO]** PR description accurately describes the changes — positive observation: The PR description is well-structured and accurately reflects the diff:
- Clearly states the motivation (scalability bottleneck, maintenance debt)
- Lists specific files changed with accurate descriptions of what changed in each
- References the relevant issue (#177)
- Includes a screenshot demonstrating the working UI
- The described changes (Sidebar dynamic map, GenericLayerNode replacing 4 custom nodes, NodePropertiesPanel rewrite, `screenToFlowPosition` fix) all match the actual diff

This is a high-quality PR description with clear before/after context.
  > **Suggestion:** No changes needed for the descriptive content itself. Only the breaking change disclosure and testing completeness need improvement (see other pr-description findings).

NamanmeetSingh:
@ivantha Thank you for catching those critical data-breaking regressions. You were completely right that shipping this without the re-hydration logic would have silently corrupted existing user canvas drafts.

I have completely resolved the review:

**Critical Fixes:**

- **[CRITICAL] Existing saved nodes lose registry data:** Added the `TYPE_MIGRATION_MAP` and `migrateNode` logic directly into the Canvas `onRestore`/load sequence. Existing drafts will now seamlessly re-hydrate their `data.registry` from the new API upon loading.

- **[CRITICAL] Conv2D parameters missing:** Added the exact `stride_rows`, `stride_cols`, and `padding` contracts back into `layer_registry.json` so the backend generation service won't break.

PR & Git Hygiene:

- Squashed all 7 commits into a single clean commit.

- Updated the PR description to explicitly check the Breaking change box and document the migration paths exactly as requested.

- Ran `prettier` and `ruff` to resolve the formatting CI failures.

The codebase is fully synced with main and ready for a final look!

ivantha:
This branch has merge conflicts with `main`. Please rebase onto the latest `main`, resolve the conflicts, and push an update. Note this also appears to depend on #178 — you may want to rebase both together.

ivantha:

PR #185, opened 2026-03-03
fix(backend): prevent Docker seed migration PermissionError on templates/json-model

dhruv1955:
## Summary
Fixes a Docker startup crash caused by Alembic seed migration writing a Keras JSON file to a path that was not writable by `appuser`.

## What changed
- Updated backend image permissions in `tensormap-backend/Dockerfile`:
  - `chown -R appuser:appuser /app`
  - `chmod -R u+w /app/templates`
- Hardened seed migration in `f4a8c2e91b37_seed_sample_project_pipeline.py`:
  - Wrapped `_generate_keras_json()` in `try/except PermissionError`
  - Emits a warning instead of crashing migration when write access is unavailable

## Why
On clean Docker startup, migration `f4a8c2e91b37` attempts to write:
`./templates/json-model/iris-classifier.json`

If container filesystem permissions are restrictive, backend enters a restart loop with:
`PermissionError: [Errno 13] Permission denied`

## Validation
- Rebuilt backend image with no cache:
  - `docker compose build --no-cache backend`
- Ran clean startup from scratch:
  - `docker compose down -v`
  - `docker compose up -d db`
  - `docker compose up -d backend`
- Verified:
  - Backend stays up (no restart loop)
  - Alembic migrations complete through `f4a8c2e91b37`
  - No `PermissionError` / `Traceback` in backend logs
  - API responds successfully:
    - `GET /api/v1/project` returns seeded `Iris Sample Project`

## Notes
- TensorFlow `cuInit` warning may appear on non-GPU machines; this is expected and unrelated to the fix.

>

### DIFF
--- a/tensormap-backend/Dockerfile
+++ b/tensormap-backend/Dockerfile
@@ -17,7 +17,8 @@ COPY --from=builder /app/.venv /app/.venv
 COPY . .
 
 RUN chown -R appuser:appuser /app \
-    && chmod -R u+w /app/templates
+    # chmod ensures appuser can write Keras JSON during seed migration (f4a8c2e91b37)
+    && chmod u+w /app/templates/json-model
 
 ENV PATH="/app/.venv/bin:$PATH"
 
--- a/tensormap-backend/migrations/versions/f4a8c2e91b37_seed_sample_project_pipeline.py
+++ b/tensormap-backend/migrations/versions/f4a8c2e91b37_seed_sample_project_pipeline.py
@@ -7,6 +7,7 @@ Create Date: 2026-02-23 22:00:00.000000
 """
 
 import json
+import logging
 import os
 import shutil
 
@@ -18,6 +19,8 @@ revision = "f4a8c2e91b37"
 down_revision = "2e66fba94864"
 branch_labels = None
 depends_on = None
+
+log = logging.getLogger(__name__)
 
 # Fixed UUIDs for idempotency
 PROJECT_UUID = "00000000-0000-4000-a000-000000000001"
@@ -118,16 +121,18 @@ def upgrade():
         )
 
     # 8. Generate Keras JSON model definition
+    # NOTE: The Dockerfile chmod ensures write access in normal operation.
+    # This guard surfaces permission failures clearly if that fix is ever bypassed.
     try:
         _generate_keras_json()
-    except PermissionError as exc:
-        import warnings
-        warnings.warn(
-            f"Could not write Keras JSON ({exc}). "
-            "The iris-classifier model JSON was not generated. "
-            "Ensure the container has write access to templates/json-model/.",
-            stacklevel=2,
+    except OSError as exc:
+        log.error(
+            "Cannot write Keras JSON to 'templates/json-model/iris-classifier.json': %s. "
+            "Ensure 'appuser' has write access to templates/json-model/. "
+            "Fix permissions and re-run migrations.",
+            exc,
         )
+        raise
 
 
 def downgrade():


tensormap-backend/migrations/versions/f4a8c2e91b37_seed_sample_project_pipeline.py
```diff
@@ -118,7 +118,16 @@ def upgrade():
...
+    try:
+        _generate_keras_json()
+    except PermissionError as exc:
+        import warnings
+        warnings.warn(
+            f"Could not write Keras JSON ({exc}). "
+            "The iris-classifier model JSON was not generated. "
+            "Ensure the container has write access to templates/json-model/.",
+            stacklevel=2,
+        )
```
ivantha: **[HIGH]** Silent failure creates irrecoverable inconsistent state

When `_generate_keras_json()` raises a `PermissionError`, the exception is caught, a warning is emitted, and the migration continues — meaning Alembic marks revision `f4a8c2e91b37` as **successfully applied** even though the filesystem artifact was never written.

This creates a **permanently inconsistent state**:
- The database already contains the seeded `Iris Sample Project` rows (inserted in steps 1–7).
- The `iris-classifier.json` file does not exist on disk.
- Alembic's version table now records the migration as complete, so it will **never be retried**.

Any downstream code that assumes the JSON file exists (e.g., model loading, API responses) will fail at runtime with a `FileNotFoundError` that is much harder to diagnose than the original `PermissionError`.

The operator has no automated way to recover without manually:
1. Reverting the migration (`alembic downgrade`).
2. Fixing permissions.
3. Re-running the upgrade.

A silent swallow of a critical side-effect inside a migration is an anti-pattern — migrations should be atomic and either fully succeed or fully fail.

**Suggestion:**
```
Do **not** catch the `PermissionError` silently. Instead, let the exception propagate so Alembic rolls back cleanly. If you want a helpful message, log it *before* re-raising:

```python
import logging
log = logging.getLogger(__name__)

# 8. Generate Keras JSON model definition
try:
    _generate_keras_json()
except PermissionError as exc:
    log.error(
        "Cannot write Keras JSON: %s. "
        "Ensure 'appuser' has write access to templates/json-model/. "
        "Fix permissions and re-run migrations.",
        exc,
    )
    raise  # Let Alembic mark the migration as FAILED
```

Alternatively, if you truly want the migration to be non-blocking (the file is optional), document this contract explicitly and add a startup-time health check that detects and reports the missing file so operators are alerted.
```

File: tensormap-backend/Dockerfile [hunk 2]
```diff
@@ -16,7 +16,8 @@ RUN groupadd -r appuser && useradd -r -g appuser appuser
 COPY --from=builder /app/.venv /app/.venv
 COPY . .
 
-RUN chown -R appuser:appuser /app
+RUN chown -R appuser:appuser /app \
+    && chmod -R u+w /app/templates
 
 ENV PATH="/app/.venv/bin:$PATH"
```
ivantha: **[LOW]** Root cause fix (Dockerfile) and defensive fallback (migration) are both applied — the combination may mask future regressions

If the Dockerfile `chown`/`chmod` is the correct and complete fix (which it likely is), then the `try/except PermissionError` in the migration becomes unreachable dead code in normal operation. This creates a maintenance hazard:

- Developers may not realize the silent-swallow behaviour exists.
- If the Dockerfile permission fix is ever reverted or broken, the migration will silently succeed (as observed in the swallow finding), hiding the regression.
- Two separate mechanisms each partially addressing the same problem makes it harder to reason about which one is actually protecting the system.

**Suggestion:**
```
Choose one of two strategies and document the decision:

**Strategy A (Preferred)**: Fix only in the Dockerfile. Remove the `try/except` entirely — the migration should remain strict. Add a comment in the Dockerfile explaining *why* the chmod is needed:
```dockerfile
# chmod ensures appuser can write Keras JSON during seed migration (f4a8c2e91b37)
RUN chown -R appuser:appuser /app \
    && chmod -R u+w /app/templates
```

**Strategy B**: Keep both, but make the migration truly defensive with re-raise (so failures still surface) and document in the migration that the Dockerfile is the primary fix:
```python
# NOTE: The Dockerfile should ensure write access. This guard is a last-resort safety net.
try:
    _generate_keras_json()
except PermissionError as exc:
    log.error("Permission denied writing Keras JSON: %s. Check Dockerfile chmod.", exc)
    raise
```
```

File: tensormap-backend/migrations/versions/f4a8c2e91b37_seed_sample_project_pipeline.py (hunk 1 above)
Comment (ivantha) [inline tensormap-backend/migrations/versions/f4a8c2e91b37_seed_sample_project_pipeline.py]: **[LOW]** No unit tests for the new error-handling path

The PR adds a new code path (the `except PermissionError` branch) with no corresponding test coverage. While testing Alembic migrations is admittedly more complex than unit-testing regular code, the new branch is entirely untested.

Without tests:
- It is impossible to verify the warning message text is correct.
- Future refactoring may accidentally remove the guard.
- The `stacklevel` bug (noted above) could have been caught by a test.

**Suggestion:**
```
Add a test that mocks `_generate_keras_json` to raise `PermissionError` and asserts that:
1. The migration does **not** propagate the exception (or does, depending on which fix strategy is chosen).
2. The correct warning/log message is emitted.

```python
import pytest
from unittest.mock import patch

def test_upgrade_handles_json_permission_error():
    with patch(
        "migrations.versions.f4a8c2e91b37_seed_sample_project_pipeline._generate_keras_json",
        side_effect=PermissionError("[Errno 13] Permission denied"),
    ):
        with pytest.warns(UserWarning, match="Could not write Keras JSON"):
            # call upgrade() with a mocked db session
            ...
```
```

File: tensormap-backend/Dockerfile (hunk 2 above)
Comment (ivantha) [inline tensormap-backend/Dockerfile]: **[INFO]** POSITIVE: Good Dockerfile layer hygiene — commands correctly combined in a single RUN

The `chown` and `chmod` commands are correctly combined into a single `RUN` instruction using `&&`. This is a Docker best practice:

- Each `RUN` instruction creates a new image layer. Splitting them would double the layer count for this step.
- More importantly, combining them avoids the common pitfall where intermediate layers (in older Docker storage drivers) capture the pre-chmod state of files, inflating image size.

This is the correct and idiomatic way to chain related filesystem operations in a Dockerfile.

**Suggestion:**
```
No change needed. This pattern is correct. Consider also verifying with `docker image inspect` that the resulting image has the expected layer count and size hasn't grown unexpectedly.
```

File: tensormap-backend/migrations/versions/f4a8c2e91b37_seed_sample_project_pipeline.py [hunk 3]
```diff
@@ -118,7 +118,16 @@ def upgrade():
...
     # 8. Generate Keras JSON model definition
-    _generate_keras_json()
+    try:
+        _generate_keras_json()
+    except PermissionError as exc:
+        import warnings
+        warnings.warn(
+            f"Could not write Keras JSON ({exc}). "
+            "The iris-classifier model JSON was not generated. "
+            "Ensure the container has write access to templates/json-model/.",
```
ivantha: **[INFO]** POSITIVE: Warning message is clear and actionable

The warning message text is well-written:
- It reports the original exception (`{exc}`) for full context.
- It names the specific artifact that was not generated (`iris-classifier model JSON`).
- It provides a concrete remediation step (`Ensure the container has write access to templates/json-model/`).

This level of detail in error messages is good practice and is better than a generic `"Failed to write file"` message.

**Suggestion:**
```
Retain this quality of messaging in whatever final error-handling strategy is chosen (logging vs warnings, re-raise vs swallow). Consider also including the full target path:
```python
f"Could not write Keras JSON to 'templates/json-model/iris-classifier.json' ({exc}). "
```
```

tensormap-backend/migrations/versions/f4a8c2e91b37_seed_sample_project_pipeline.py
dhruv1955: Fixed — removed the silent swallow. Exception is now logged with full 
context and re-raised so Alembic marks the migration as FAILED, 
preventing inconsistent state.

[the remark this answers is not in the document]
dhruv1955: Fixed — broadened from PermissionError to OSError, which covers 
PermissionError, FileNotFoundError, disk-full, and IsADirectoryError 
as all are OSError subclasses.

PR #186, opened 2026-03-03
feat(backend): improve model generation error handling and validation

Grayking1905:
**# Improved Error Handling for Model Validation**

**## Problem Description**

When a user designs a neural network on the TensorMap canvas and clicks "Validate," the backend's `model_generation()` function could previously fail with cryptic TensorFlow errors or stack traces if given invalid graph definitions. This made it difficult for users to know what went wrong (e.g. disconnected nodes, missing inputs, incompatible shapes, or invalid parameters).

Fix - #146 

**## Changes Made**

This PR implements structured, upfront graph validation to protect the Keras compilation process and provide user-friendly error messages directly to the client.

- **Pre-build Graph Validation**
    - Added `_validate_graph()` to `app/services/model_generation.py` to check for missing input nodes, unknown layer types, and non-numeric parameters _before_ TensorFlow tries to parse them.
    - Implemented a graph search algorithm to identify and list disconnected/orphan nodes.
- **Custom Exception**
    - Introduced `ModelValidationError` to securely carry user-facing error messages to the API layer without leaking paths.
- **API Layer Support**
    - Updated `model_validate_service` and `model_save_service` in `app/services/deep_learning.py` to catch `ModelValidationError` and immediately return a clean `400 Bad Request` response holding the API's standard `{"success": false, "message": "..."}` structure.
    - Expanded `app/shared/errors.py` fallback messages for remaining low-level TensorFlow exceptions (like invalid `Concatenate` branches).
- **Testing**
    - Thoroughly expanded `tests/test_model_generation.py` with full coverage for all error pathways and validation strategies. All 35 tests pass smoothly with 0 lint errors system-wide.

**## Acceptance Criteria Met**

- [x] Disconnected graph produces: "Disconnected graph: N node(s) are not connected to the Input layer..."
- [x] Missing Input node produces: "No Input layer found..."
- [x] Shape mismatch (e.g., Dense after Conv2D) produces a message explaining the need for Flatten
- [x] Invalid parameter values (e.g., non-numeric units) produce a message naming the specific node and parameter
- [x] Unknown node types produce a message listing supported types
- [x] All error messages are user-friendly and suggest how to fix the problem
- [x] Error responses include the standard `{"success": false, "message": "..."}` format
- [x] Generic TensorFlow errors are caught and wrapped with context
- [x] Valid models continue to validate successfully (no regression)
- [x] Error messages do not expose internal file paths or stack traces

### DIFF
--- a/tensormap-backend/app/services/model_generation.py
+++ b/tensormap-backend/app/services/model_generation.py
@@ -9,6 +9,124 @@ from app.shared.logging_config import get_logger
 
 logger = get_logger(__name__)
 
+# Node types that model_generation() knows how to build.
+SUPPORTED_NODE_TYPES = {"custominput", "customdense", "customflatten", "customconv"}
+
+
+class ModelValidationError(ValueError):
+    """Raised when the ReactFlow graph fails pre-build validation.
+
+    Carries a ``user_message`` that is safe to return directly to the client
+    (no internal paths, stack traces, or TensorFlow class names).
+    """
+
+    def __init__(self, user_message: str) -> None:
+        super().__init__(user_message)
+        self.user_message = user_message
+
+
+def _validate_graph(nodes: list[dict], edges: list[dict]) -> None:
+    """Run pre-build checks on the ReactFlow graph and raise ModelValidationError on failure.
+
+    Checks performed (in order):
+    1. At least one ``custominput`` node is present.
+    2. Every node type is in SUPPORTED_NODE_TYPES.
+    3. All layer parameters that must be numeric are valid integers.
+    4. Every non-input node is reachable from an input node (disconnected graph).
+    """
+    # ------------------------------------------------------------------
+    # 1.  Must have at least one Input layer
+    # ------------------------------------------------------------------
+    input_node_ids = {n["id"] for n in nodes if n["type"] == "custominput"}
+    if not input_node_ids:
+        raise ModelValidationError(
+            "No Input layer found. Add an Input node to your canvas and connect it "
+            "to the rest of the network before validating."
+        )
+
+    # ------------------------------------------------------------------
+    # 2. Unknown node types
+    # ------------------------------------------------------------------
+    unknown = [n for n in nodes if n["type"] not in SUPPORTED_NODE_TYPES]
+    if unknown:
+        labels = ", ".join(f'"{n["type"]}"' for n in unknown)
+        supported = ", ".join(sorted(SUPPORTED_NODE_TYPES - {"custominput"}))
+        raise ModelValidationError(
+            f"Unknown layer type(s): {labels}. "
+            f"Supported types are: custominput, {supported}."
+        )
+
+    # ------------------------------------------------------------------
+    # 3. Parameter type validation
+    # ------------------------------------------------------------------
+    for node in nodes:
+        node_id = node["id"]
+        params = node["data"].get("params", {})
+        node_type = node["type"]
+
+        int_fields: dict[str, str] = {}
+        if node_type == "customdense":
+            int_fields = {"units": "Units"}
+        elif node_type == "customconv":
+            int_fields = {
+                "filter": "Filters",
+                "kernelX": "Kernel width",
+                "kernelY": "Kernel height",
+                "strideX": "Stride width",
+                "strideY": "Stride height",
+            }
+        elif node_type == "custominput":
+            for i in range(1, 4):
+                key = f"dim-{i}"
+                val = params.get(key)
+                if val is not None and val != "" and val != 0:
+                    try:
+                        int(val)
+                    except (TypeError, ValueError):
+                        raise ModelValidationError(
+                            f'Node "{node_id}": dimension "{key}" must be a positive integer, '
+                            f'got "{val}". Fix the Input layer parameters and try again.'
+                        ) from None
+
+        for field, label in int_fields.items():
+            val = params.get(field)
+            try:
+                int(val)
+            except (TypeError, ValueError):
+                raise ModelValidationError(
+                    f'Node "{node_id}": "{label}" must be a positive integer, '
+                    f'got "{val}". Fix the parameter and try again.'
+                ) from None
+
+    # ------------------------------------------------------------------
+    # 4. Disconnected graph — BFS from all input nodes
+    # ------------------------------------------------------------------
+    adjacency: dict[str, list[str]] = defaultdict(list)
+    for edge in edges:
+        adjacency[edge["source"]].append(edge["target"])
+        adjacency[edge["target"]].append(edge["source"])
+
+    reachable: set[str] = set()
+    queue = list(input_node_ids)
+    reachable.update(input_node_ids)
+
+    while queue:
+        current = queue.pop(0)
+        for neighbour in adjacency.get(current, []):
+            if neighbour not in reachable:
+                reachable.add(neighbour)
+                queue.append(neighbour)
+
+    all_ids = {n["id"] for n in nodes}
+    disconnected = all_ids - reachable
+    if disconnected:
+        count = len(disconnected)
+        raise ModelValidationError(
+            f"Disconnected graph: {count} node(s) are not connected to the Input layer "
+            f"({', '.join(sorted(disconnected))}). Connect every node to the network "
+            "before validating."
+        )
+
 
 def model_generation(model_params: dict) -> dict:
     """Transform ReactFlow nodes and edges into a Keras functional-API JSON structure.
@@ -16,24 +134,34 @@ def model_generation(model_params: dict) -> dict:
     Builds the model programmatically using the Keras API, then serializes
     it via ``model.to_json()`` so the output always matches the installed
     Keras version's expected format.
+
+    Raises:
+        ModelValidationError: Structured, user-friendly error for any graph or
+            parameter problem detected *before* or *during* the Keras build.
     """
-    logger.debug("Generating model from %d nodes, %d edges", len(model_params["nodes"]), len(model_params["edges"]))
+    nodes: list[dict] = model_params["nodes"]
+    edges: list[dict] = model_params["edges"]
+
+    logger.debug("Generating model from %d nodes, %d edges", len(nodes), len(edges))
+
+    # -- Pre-build validation (fast-path, no TF involved) ----------------
+    _validate_graph(nodes, edges)
 
     # Build adjacency maps
-    source_to_targets = defaultdict(list)
-    target_to_sources = defaultdict(list)
-    for edge in model_params["edges"]:
+    source_to_targets: dict[str, list[str]] = defaultdict(list)
+    target_to_sources: dict[str, list[str]] = defaultdict(list)
+    for edge in edges:
         source_to_targets[edge["source"]].append(edge["target"])
         target_to_sources[edge["target"]].append(edge["source"])
 
-    nodes_by_id = {node["id"]: node for node in model_params["nodes"]}
+    nodes_by_id = {node["id"]: node for node in nodes}
 
     # BFS from input nodes to build Keras layers in topological order
-    keras_tensors = {}
-    visited = set()
-    queue = []
+    keras_tensors: dict[str, object] = {}
+    visited: set[str] = set()
+    queue: list[str] = []
 
-    for node in model_params["nodes"]:
+    for node in nodes:
         if node["type"] == "custominput":
             dims = [int(node["data"]["params"].get(f"dim-{i + 1}", 0) or 0) for i in range(3)]
             dims = [d for d in dims if d != 0]
@@ -46,7 +174,6 @@ def model_generation(model_params: dict) -> dict:
         for target_id in source_to_targets.get(current_id, []):
             if target_id in visited:
                 continue
-            # Check if all sources of this target have been visited
             all_sources = target_to_sources.get(target_id, [])
             if not all(src in visited for src in all_sources):
                 continue
@@ -54,7 +181,6 @@ def model_generation(model_params: dict) -> dict:
             visited.add(target_id)
             queue.append(target_id)
 
-            # Collect input tensors for this node
             source_tensors = [keras_tensors[src] for src in all_sources]
             if len(source_tensors) > 1:
                 input_tensor = tf.keras.layers.Concatenate(axis=-1)(source_tensors)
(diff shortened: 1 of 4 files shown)


ivantha (commented):
## ◎ Argus Review

This PR successfully implements structured pre-build validation for the TensorMap model builder, replacing opaque TensorFlow stack traces with user-friendly error messages. The architecture is sound — the `ModelValidationError` / `_validate_graph` separation is clean, the acceptance-criteria test suite is comprehensive, and the `ValueError` inheritance preserves backward compatibility nicely. The primary concerns are a potential unhandled `KeyError` (→ HTTP 500) if a node is missing its `data` key, a generic error branch in `_wrap_build_error` that contradicts the PR's own no-leak acceptance criterion, and numeric validation that claims to enforce positivity but only checks parseability. Several lower-severity code-quality items (duplicate BFS logic, O(n) queue pop, redundant exception handlers) are worth addressing before merge.

tensormap-backend/app/services/model_generation.py
```diff
@@ -9,31 +9,159 @@
...
+        for field, label in int_fields.items():
+            val = params.get(field)
+            try:
+                int(val)
+            except (TypeError, ValueError):
+                raise ModelValidationError(
+                    f'Node "{node_id}": "{label}" must be a positive integer, '
+                    f'got "{val}". Fix the parameter and try again.'
+                ) from None
+
```
ivantha: **[MEDIUM]** Parameter validation claims "positive integer" but accepts zero and negative values

The error messages in `_validate_graph` state *"must be a positive integer"* for `units`, filters, kernel sizes, strides, and input dimensions. However, the actual check is:
```python
try:
    int(val)
except (TypeError, ValueError):
    raise ModelValidationError(...)
```
This only verifies parseability — it does **not** reject `0`, `-1`, or other non-positive numbers. A `Dense` layer with `units=0` or `Conv2D` with `filters=-3` will pass `_validate_graph` and then fail inside TensorFlow with a cryptic message, bypassing the validation layer introduced in this PR.

**Suggestion:**
```
After `int(val)`, add a positivity assertion:
```python
for field, label in int_fields.items():
    val = params.get(field)
    try:
        parsed = int(val)
    except (TypeError, ValueError):
        raise ModelValidationError(
            f'Node "{node_id}": "{label}" must be a positive integer, '
            f'got "{val}". Fix the parameter and try again.'
        ) from None
    if parsed <= 0:
        raise ModelValidationError(
            f'Node "{node_id}": "{label}" must be a positive integer, '
            f'got {parsed}. Fix the parameter and try again.'
        )
```
```

File: tensormap-backend/app/services/model_generation.py [hunk 2]
```diff
@@ -9,31 +9,159 @@
...
+    # ------------------------------------------------------------------
+    # 3. Parameter type validation
+    # ------------------------------------------------------------------
+    for node in nodes:
+        node_id = node["id"]
+        params = node["data"].get("params", {})
+        node_type = node["type"]
+
+        int_fields: dict[str, str] = {}
+        if node_type == "customdense":
```
ivantha: **[MEDIUM]** `custominput` dim-validation zero-check is type-inconsistent (`val != 0` vs. string `"0"`)

The guard for skipping unset Input dimensions is:
```python
if val is not None and val != "" and val != 0:
    try:
        int(val)
    ...
```
`val != 0` short-circuits only when `val` is the integer `0`. Values arriving from the frontend JSON are almost always **strings** (e.g., `"0"`), and `"0" != 0` is `True` in Python 3, so string `"0"` falls through to `int()` and is silently accepted as a valid dimension. An integer `0` is silently skipped. This asymmetry is confusing, and a dimension of `0` is invalid in Keras (`InputLayer` with shape `(0,)` raises a build error) so it should arguably be rejected rather than silently skipped.

**Suggestion:**
```
Normalise the check:
```python
if val is not None and str(val).strip() not in ("", "0"):
    try:
        parsed = int(val)
    except (TypeError, ValueError):
        raise ModelValidationError(...) from None
    if parsed <= 0:
        raise ModelValidationError(
            f'Node "{node_id}": dimension "{key}" must be a positive integer, '
            f'got "{val}".'
        )
```
Or add a comment explaining why `0` is treated as "not set" if that is intentional.
```

File: tensormap-backend/app/services/model_generation.py [hunk 3]
```diff
@@ -9,31 +9,159 @@
...
+
+    all_ids = {n["id"] for n in nodes}
+    disconnected = all_ids - reachable
+    if disconnected:
+        count = len(disconnected)
+        raise ModelValidationError(
+            f"Disconnected graph: {count} node(s) are not connected to the Input layer "
+            f"({', '.join(sorted(disconnected))}). Connect every node to the network "
+            "before validating."
+        )
```
ivantha: **[MEDIUM]** `queue.pop(0)` is O(n) in both BFS implementations — use `collections.deque`

Both the new `_validate_graph` BFS and the pre-existing `model_generation` BFS use a plain `list` as a FIFO queue and call `queue.pop(0)`. `list.pop(0)` shifts all remaining elements and is O(n), making the overall BFS O(n²) in the number of nodes. `collections.deque` is already imported (since `defaultdict` is from `collections`), and `deque.popleft()` is O(1).

**Suggestion:**
```
```python
from collections import defaultdict, deque

# In _validate_graph:
queue: deque[str] = deque(input_node_ids)
reachable.update(input_node_ids)
while queue:
    current = queue.popleft()  # O(1)
    ...

# In model_generation:
queue: deque[str] = deque()
...
current_id = queue.popleft()  # O(1)
```
```

File: tensormap-backend/app/services/model_generation.py [hunk 4]
```diff
@@ -46,33 +174,101 @@ def model_generation(model_params: dict) -> dict:
...
                 input_tensor = source_tensors[0]
 
             node = nodes_by_id[target_id]
-            keras_tensors[target_id] = _build_layer(node, input_tensor)
+            try:
+                keras_tensors[target_id] = _build_layer(node, input_tensor)
+            except ModelValidationError:
+                raise
+            except (ValueError, TypeError) as exc:
+                raise ModelValidationError(_wrap_build_error(node, exc)) from exc
```
ivantha: **[MEDIUM]** Duplicate adjacency-map construction and BFS traversal between `_validate_graph` and `model_generation`

`_validate_graph` builds its own undirected `adjacency` dict and performs a full BFS to detect disconnected nodes. `model_generation` then builds two separate adjacency dicts (`source_to_targets`, `target_to_sources`) and performs another BFS for topological ordering. Edge iteration and graph-traversal code is duplicated across both functions — any future change to edge representation (e.g., renaming `"source"` / `"target"` keys) must be made in two places.

**Suggestion:**
```
Extract adjacency construction into a shared helper and pass the result into `_validate_graph`:
```python
def _build_adjacency(
    edges: list[dict]
) -> tuple[dict[str, list[str]], dict[str, list[str]]]:
    source_to_targets: dict[str, list[str]] = defaultdict(list)
    target_to_sources: dict[str, list[str]] = defaultdict(list)
    for edge in edges:
        source_to_targets[edge["source"]].append(edge["target"])
        target_to_sources[edge["target"]].append(edge["source"])
    return source_to_targets, target_to_sources
```
Then call once in `model_generation` and pass the maps into `_validate_graph`.
```

Comment (ivantha) [inline tensormap-backend/app/services/model_generation.py]: **[LOW]** Redundant `except (ValueError, TypeError)` handler — completely absorbed by subsequent `except Exception`

Inside the BFS loop, two consecutive exception clauses handle `_build_layer` failures identically:
```python
except ModelValidationError:
    raise
except (ValueError, TypeError) as exc:
    raise ModelValidationError(_wrap_build_error(node, exc)) from exc
except Exception as exc:  # noqa: BLE001
    raise ModelValidationError(_wrap_build_error(node, exc)) from exc
```
Both the `(ValueError, TypeError)` and the `Exception` branches call `_wrap_build_error` with the same arguments. Since `ValueError` and `TypeError` are subclasses of `Exception`, the first branch is entirely shadowed by the second and provides no distinct behaviour.

**Suggestion:**
```
Collapse to two handlers:
```python
except ModelValidationError:
    raise
except Exception as exc:  # noqa: BLE001
    raise ModelValidationError(_wrap_build_error(node, exc)) from exc
```
If distinguishing TF value errors from unexpected crashes for logging is desired, add a `logger.debug` call in the `(ValueError, TypeError)` branch before re-raising.
```

File: tensormap-backend/app/services/model_generation.py [hunk 5]
```diff
@@ -9,31 +9,159 @@
...
+    unknown = [n for n in nodes if n["type"] not in SUPPORTED_NODE_TYPES]
+    if unknown:
+        labels = ", ".join(f'"{n["type"]}"' for n in unknown)
+        supported = ", ".join(sorted(SUPPORTED_NODE_TYPES - {"custominput"}))
+        raise ModelValidationError(
+            f"Unknown layer type(s): {labels}. "
+            f"Supported types are: custominput, {supported}."
+        )
+
+    # ------------------------------------------------------------------
```
ivantha: **[LOW]** Inconsistent "unknown node type" error message format between `_validate_graph` and `_build_layer`

`_validate_graph` raises: `"Unknown layer type(s): \"customunknown\". Supported types are: custominput, ..."` while `_build_layer` raises: `'Unknown layer type \"customunknown\" on node \"mynode\". Supported types are: ...'`. The phrasing differs (`"layer type(s)"` vs `"layer type"`), one includes the node ID and the other does not, and they have different sentence structures. While in production `_build_layer`'s branch is now unreachable for unknown types (since `_validate_graph` fires first), tests call `_build_layer` directly and the inconsistency already surfaced with the renamed test.

**Suggestion:**
```
Either remove the `else` branch in `_build_layer` with an `assert False, "unreachable: unknown types filtered by _validate_graph"` guard, or extract a shared helper:
```python
def _unknown_type_message(node_id: str, node_type: str) -> str:
    supported = ", ".join(sorted(SUPPORTED_NODE_TYPES))
    return (
        f'Unknown layer type "{node_type}" on node "{node_id}". '
        f'Supported types are: {supported}.'
    )
```
```

Comment (ivantha) [inline tensormap-backend/app/services/model_generation.py]: **[LOW]** Hardcoded `custominput` prefix in unknown-type message is fragile

```python
supported = ", ".join(sorted(SUPPORTED_NODE_TYPES - {"custominput"}))
raise ModelValidationError(
    f"Unknown layer type(s): {labels}. "
    f"Supported types are: custominput, {supported}."
)
```
The string `"custominput"` is hardcoded as a literal prefix in the f-string rather than being derived from `SUPPORTED_NODE_TYPES`. If the constant is ever updated (type renamed or removed), this message would list a stale type name.

**Suggestion:**
```
```python
supported = ", ".join(sorted(SUPPORTED_NODE_TYPES))
raise ModelValidationError(
    f"Unknown layer type(s): {labels}. "
    f"Supported types are: {supported}."
)
```
```

File: tensormap-backend/tests/test_model_generation.py [hunk 6]
```diff
@@ -263,3 +394,120 @@ def test_multi_dim_input_shape(self):
...
+        assert "No Input layer found" in exc_info.value.user_message
+
+    def test_disconnected_graph_message_contains_count(self):
+        """Disconnected orphan node → error with node count and node ID."""
+        params = {
+            "nodes": [
+                _input_node("i1", [10]),
+                _dense_node("d1"),
+                _dense_node("orphan"),
+            ],
```
ivantha: **[LOW]** `test_single_input_no_edges_raises` accepts both exception types — unclear expectation

```python
with pytest.raises((ModelValidationError, ValueError)):
    model_generation(params)
```
This test passes regardless of which exception type escapes, providing no signal about whether the PR's migration to `ModelValidationError` is complete for this case. The earlier version of this test (`test_single_input_no_edges`) explicitly expected `ValueError` with a comment explaining that Keras 3's Functional API rejects an output-equals-input topology. That contract has been silently widened.

**Suggestion:**
```
Decide the expected contract and pin it:
```python
# If wrapping is intended:
with pytest.raises(ModelValidationError, match="output cannot be the same tensor"):
    model_generation(params)

# If Keras ValueError is acceptable:
with pytest.raises(ValueError):
    model_generation(params)
```
If the exception type genuinely varies by Keras version, document that with a comment and consider parametrising by version.
```

**Review by ivantha — COMMENTED** [review]
## ◎ Argus Review

This PR delivers a well-structured and clearly motivated improvement to model validation error handling. The introduction of `ModelValidationError`, the `_validate_graph` pre-flight checks, and the comprehensive test suite are all solid work. However, there is one **high-severity logic bug**: the disconnected-graph BFS in `_validate_graph` uses an undirected adjacency map, causing false negatives for nodes that have an outgoing edge into the reachable subgraph but no incoming path from Input — such nodes will be incorrectly classified as reachable, defeating the check. There are also medium-severity concerns around unguarded `node["data"]` access that could surface raw `KeyError`s, an overly broad `"Invalid argument"` fallback key in `errors.py`, and fragile version-sensitive TF error string matching in `_wrap_build_error`. The test suite is thorough but two test gaps weaken the safety net: `test_single_input_no_edges_raises` accepts either exception type (masking potential regressions), and the path-leak test does not cover Unix-style forward-slash paths nor exercise the actual path-scrubbing codepath.

**Suggested Version Bump:** `minor`

### ⚠️ Breaking Changes

- **[INFO]** _build_layer no longer raises ValueError — raises ModelValidationError instead: **Previous behavior**: `_build_layer()` raised a plain `ValueError` for unknown node types:
```python
raise ValueError(f"Unknown node type: {node_type}")
```

**New behavior**: Raises `ModelValidationError` (a subclass of `ValueError`) with a richer message.

**Who is affected**: Any direct caller of `_build_layer()` that catches `ValueError` specifically with `except ValueError as e` and inspects `str(e)` for the old substring `"Unknown node type"`. The old test `pytest.raises(ValueError, match="Unknown node type")` has been updated in this PR, but any external code or out-of-tree tests that relied on the old exception message pattern would silently continue to catch the exception (since `ModelValidationError` is a `ValueError`) but would fail message-match assertions.

**Migration path**: Replace `except ValueError` catches that inspect the old message with `except ModelValidationError` catches that inspect `.user_message`, or simply rely on the backward-compatible `ValueError` base class.
  > **Suggestion:** No immediate migration required for code catching `ValueError` generically. For code matching the error string `"Unknown node type"`, update to match `"Unknown layer type"` (the new message wording).
- **[LOW]** Client-visible error message strings changed in errors.py: **Previous behavior**: `errors.py` returned the following client-visible strings:
- `"Incompatible inputs for multi-input layers."`
- `"Model name already exists"` (no trailing period)
- `"MODEL RUNNING FAILED : Incompatible shapes for input or output layers"`
- `"Error in Model values"`

**New behavior**: All four messages are replaced with more descriptive alternatives.

**Who is affected**: Any client application (frontend, integration test, or API consumer) that performs **exact string comparison** on the `message` field of the `{"success": false, "message": "..."}` response. The old `"Model name already exists"` → `"Model name already exists."` change (period added) is particularly subtle.

**Migration path**: Clients should test for `success === false` rather than matching exact message strings. If exact matching is required, update the expected strings to the new values.
  > **Suggestion:** Document the message changes in the PR description or a CHANGELOG entry to signal to frontend/integration-test authors that string-matching assertions need to be updated.


### PR Description Issues

- **[INFO]** PR description is thorough and well-structured — minor gap: no mention of the undirected BFS choice: The PR description is excellent overall: it clearly states the problem, enumerates every change, maps them to acceptance criteria, and confirms the test count. The acceptance criteria checkboxes are all filled, and there are no template placeholders.

One small gap: the description does not explain the **algorithmic choice** made in the disconnected-graph check (undirected BFS). This is the most architecturally significant decision in the PR and is also where the current bug (false negatives) originates. A note explaining why an undirected vs directed traversal was chosen — or acknowledging the trade-off — would help reviewers evaluate correctness quickly.

Another minor gap: the description says `_validate_graph` is added to `app/services/model_generation.py` but does not mention that `_wrap_build_error` and `_wrap_model_build_error` were also added as new private helpers. These are visible in the diff but not called out explicitly.
  > **Suggestion:** Add a brief section to "Changes Made" noting:
- `_wrap_build_error()` and `_wrap_model_build_error()` as new private helpers responsible for post-hoc error translation.
- A note clarifying that the disconnected-graph BFS uses undirected traversal and why (or confirming it intentionally uses directed traversal, which is what the test expectations imply).

ivantha:
CI is failing on **Backend Lint**. Please run `uv run ruff check . && uv run ruff format --check .` inside `tensormap-backend/`, fix the reported issues, and push an update.

ivantha:
Thank you for the model generation error handling improvements! The error-handling landscape in this area has been significantly improved by merged PRs #260, #226, and #183. This PR conflicts with main and has CI failures; the core improvements are now covered elsewhere. We appreciate the work on #186 and encourage you to contribute to other open issues.

PR #194, opened 2026-03-05
fix: add null checks in model_run.py to prevent AttributeError

### LINKED ISSUE #193: [Bug] model_run.py crashes with AttributeError when file_id or model_name not found in DB
## Describe the bug
Two functions in `app/services/model_run.py` perform DB lookups without null checks, causing unhandled `AttributeError` crashes:

**1. `_helper_generate_file_location`**
```python
file = db.exec(select(DataFile).where(DataFile.id == file_id)).first()
if file.file_type == "zip":  # crashes if file is None
```

**2. `_run`**
```python
model_configs = db.exec(...).first()
if model_configs.model_type == ...:  # crashes if model_configs is None
```

## Steps to Reproduce
1. Call `POST /api/v1/model/run/<model_name>` with a model_name that exists in DB but whose `file_id` references a deleted file
2. Observe: `AttributeError: 'NoneType' object has no attribute 'file_type'`

## Expected behavior
Both functions should return a clear 404 error message when the DB record is not found, instead of crashing with AttributeError.

## Suggested Fix
```python
def _helper_generate_file_location(db: Session, file_id) -> str:
    file = db.exec(select(DataFile).where(DataFile.id == file_id)).first()
    if file is None:
        raise ValueError(f"File with id {file_id} not found in database")
    ...
```

VishakBaddur:
## Summary
Fixes #193

Two functions in `app/services/model_run.py` performed DB lookups without null checks, causing unhandled `AttributeError` crashes.

## Changes

**`_helper_generate_file_location`** - added null check after DB lookup:
- Before: `file.file_type` crashed with `AttributeError` if file_id 
  referenced a deleted or non-existent file
- After: raises `ValueError` with a clear message

**`_run`** - added null check after model lookup:
- Before: `model_configs.model_type` crashed with `AttributeError` if model_name not found in DB
- After: raises `ValueError` with a clear message

Both errors are caught by the caller `model_run()` which already has a try/except that emits the error via Socket.IO to the frontend.


ivantha:
## ◎ Argus Review

This PR correctly identifies and patches two `AttributeError` crash sites in `model_run.py` by adding null guards after database lookups and surfacing them as descriptive `ValueError` exceptions. The fix is minimal, targeted, and leverages the existing exception-handling boundary in `model_run()`. However, the fix is incomplete: `_run` contains several additional DB lookups immediately after the patched area that are still missing null checks, meaning the same class of crash remains possible. No automated tests accompany the change, leaving the new error paths unverified.

### Findings (9)

#### 🟠 Incomplete null-check coverage in `_run` — subsequent DB lookups still unguarded
**Severity:** high | **Category:** bug | **File:** `tensormap-backend/app/services/model_run.py:124`

The PR adds a null check for `model_configs`, but the very next lines inside `_run` perform additional `db.exec(...).first()` calls whose results are immediately dereferenced without any null guard:

```python
image_properties = db.exec(select(ImageProperties).where(ImageProperties.id == model_configs.file_id)).first()
# image_properties used below with no null check
```

If a row in `ImageProperties`, `TextProperties`, or any other queried table is missing (e.g., deleted record, FK violation, partial insert), the same `AttributeError` crash this PR is trying to fix will still occur. The fix therefore only partially solves the problem — it guards the entry point of `_run` but leaves the body unprotected.

**Suggestion:**
```
Apply the same pattern to every `.first()` result that is subsequently dereferenced:

```python
image_properties = db.exec(
    select(ImageProperties).where(ImageProperties.id == model_configs.file_id)
).first()
if image_properties is None:
    raise ValueError(
        f"ImageProperties for file_id {model_configs.file_id} not found in database"
    )
```

Repeat for every other `.first()` call whose return value is not immediately checked.
```

---

#### 🟡 No unit tests for the new null-check error paths
**Severity:** medium | **Category:** testing | **File:** `tensormap-backend/app/services/model_run.py:91`

The PR description references a specific bug (issue #193) and describes exact failure conditions (`file_id` referencing a deleted file, `model_name` not found). Despite this precise understanding of the failure modes, no automated tests are added to exercise the new `ValueError` branches.

Without tests:
- Future refactors could silently remove the null checks.
- The claim that `model_run()` catches these exceptions and emits them via Socket.IO is unverified.
- Regressions of issue #193 would go undetected.

**Suggestion:**
```
Add pytest unit tests that mock `db.exec(...).first()` to return `None` and assert `ValueError` is raised:

```python
def test_helper_generate_file_location_raises_when_file_not_found(mock_db):
    mock_db.exec.return_value.first.return_value = None
    with pytest.raises(ValueError, match="File with id 999 not found in database"):
        _helper_generate_file_location(mock_db, file_id=999)

def test_run_raises_when_model_not_found(mock_db):
    mock_db.exec.return_value.first.return_value = None
    with pytest.raises(ValueError, match="Model 'ghost_model' not found in database"):
        _run("ghost_model", mock_db)
```
```

---

#### 🟡 Internal DB identifiers leaked in error messages forwarded to the frontend
**Severity:** medium | **Category:** security | **File:** `tensormap-backend/app/services/model_run.py:92`

The error message `f"File with id {file_id} not found in database"` embeds an internal primary key which, per the PR description, is forwarded directly to the frontend via Socket.IO in `model_run()`'s `except` block.

Exposing internal database IDs to clients:
- Reveals the auto-increment sequence or UUID scheme, aiding enumeration attacks.
- Confirms which IDs exist vs. do not exist (oracle attack surface).
- Violates the principle of keeping internal implementation details server-side.

While this appears to be an authenticated internal tool, the pattern is worth correcting.

**Suggestion:**
```
Log the full details server-side and emit a sanitised message to the client:

```python
import logging
logger = logging.getLogger(__name__)

if file is None:
    logger.error("File lookup failed: id=%s not found", file_id)
    raise ValueError("Requested file could not be located. Please check your configuration.")
```
```

---

#### 🔵 Missing type annotation on `file_id` parameter
**Severity:** low | **Category:** maintainability | **File:** `tensormap-backend/app/services/model_run.py:88`

The function signature `def _helper_generate_file_location(db: Session, file_id) -> str:` omits the type for `file_id`. This is inconsistent with `_run` which fully annotates its parameters.

The missing annotation:
- Makes it ambiguous whether `file_id` is `int`, `uuid.UUID`, or `str`.
- Prevents static type checkers (mypy, pyright) from catching mismatched call sites.
- Is especially relevant now that invalid IDs trigger a `ValueError` rather than a silent `None`.

**Suggestion:**
```
Add the appropriate type annotation (adjust to `uuid.UUID` if applicable):

```python
def _helper_generate_file_location(db: Session, file_id: int) -> str:
```
```

---

#### 🔵 Docstring not updated to document the new exception behaviour
**Severity:** low | **Category:** documentation | **File:** `tensormap-backend/app/services/model_run.py:88`

The docstring for `_helper_generate_file_location` reads:

```
"""Resolve the on-disk path for a dataset file by its DB ID."""
```

It does not document that the function now **raises `ValueError`** when the file is not found. Callers and future maintainers have no way to know about this contract from the docstring alone.

**Suggestion:**
```
Update the docstring to include the raised exception:

```python
def _helper_generate_file_location(db: Session, file_id: int) -> str:
    """
    Resolve the on-disk path for a dataset file by its DB ID.

    Raises:
        ValueError: If no DataFile with the given file_id exists in the database.
    """
```
```

---

#### 🔵 Path construction uses fragile string concatenation instead of `os.path.join`
**Severity:** low | **Category:** maintainability | **File:** `tensormap-backend/app/services/model_run.py:94`

The existing path construction logic uses raw string concatenation:

```python
return upload_folder + "/" + file.file_name
return upload_folder + "/" + file.file_name + "." + file.file_type
```

This is error-prone: if `upload_folder` already ends with `/`, the resulting path will contain a double slash. The hardcoded `/` separator is also non-portable. While this is pre-existing code, the null-check changes touch these exact lines and this is a good moment to flag it.

**Suggestion:**
```
Use `pathlib.Path` for robust, cross-platform path construction:

```python
from pathlib import Path

base = Path(get_settings().upload_folder)
if file.file_type == "zip":
    return str(base / file.file_name)
return str(base / f"{file.file_name}.{file.file_type}")
```
```

---

#### 🔵 Consider a dedicated domain exception hierarchy instead of bare `ValueError`
**Severity:** low | **Category:** maintainability | **File:** `tensormap-backend/app/services/model_run.py:91`

`ValueError` has a broad semantic meaning in Python. Using it for 'entity not found in database' conflates two unrelated failure modes: programming errors (wrong value passed) vs. data-integrity errors (entity deleted between UI action and execution).

As the codebase grows, the flat `except Exception` in `model_run()` treating all `ValueError`s uniformly makes fine-grained error reporting to the frontend harder and could obscure unrelated bugs.

**Suggestion:**
```
Introduce lightweight custom exceptions:

```python
# app/exceptions.py
class EntityNotFoundError(LookupError):
    """Raised when a required database entity is missing."""

class FileNotFoundInDBError(EntityNotFoundError): ...
class ModelNotFoundError(EntityNotFoundError): ...
```

The Socket.IO emitter can then distinguish between recoverable 'entity not found' errors and unexpected runtime errors.
```

---

#### ℹ️ Asymmetric file-path logic for zip vs. non-zip files (pre-existing, visible in diff context)
**Severity:** info | **Category:** bug | **File:** `tensormap-backend/app/services/model_run.py:94`

The path construction distinguishes zip files:

```python
if file.file_type == "zip":
    return upload_folder + "/" + file.file_name          # no extension appended
return upload_folder + "/" + file.file_name + "." + file.file_type  # extension appended
```

This implies `file_name` for zip files already includes the `.zip` extension, while other types store only the stem. This implicit convention is undocumented and easy to violate. If a zip file is ever stored without its extension in `file_name`, the function will silently return an incorrect path.

**Suggestion:**
```
Document the convention explicitly in the model, or enforce it at the DB layer. Alternatively, unify the logic to always store the stem and derive the full path uniformly:

```python
return str(Path(upload_folder) / f"{file.file_name}.{file.file_type}")
```
and migrate existing zip records to strip `.zip` from `file_name`.
```

---

#### ℹ️ ✅ Positive: correct exception type and descriptive messages
**Severity:** info | **Category:** maintainability | **File:** `tensormap-backend/app/services/model_run.py:91`

Both null checks raise `ValueError` (appropriate for 'entity not found given valid input') rather than a bare `AttributeError` or a generic `Exception`. The messages include the offending identifier (`file_id`, `model_name`), making them immediately actionable for debugging. This follows good defensive-programming practice and is consistent with Python conventions.

**Suggestion:**
```
No change needed here — this is a good pattern to replicate for the remaining unguarded lookups elsewhere in `_run`.
```

---


*◎ Generated by [Argus](https://github.com) — AI-Powered PR Review*

ivantha:

ivantha:
Thank you for the model_run.py null-check fix (#193)! This work has been superseded by #269 which addresses the same AttributeError scenarios across model_run and other services comprehensively. We appreciate the targeted bug fix.

PR #196, opened 2026-03-05
fix: add null checks and KeyError guard in code_generation.py

### LINKED ISSUE #195: [Bug] _map_template raises KeyError on an unknown model_type
**Difficulty:** Easy
**Area:** backend

## Description
The null-check half of this report has landed: `generate_code()` now raises `CodeGenerationError` when the model or its dataset file is missing (`code_generation.py:37-42`), and `get_code_service()` turns that into a 404. What survives is the template lookup. `_map_template()` indexes its `options` dict directly, so any `model_type` outside `ProblemType` escapes as a bare `KeyError`, including `None`, which `ModelBasic.model_type` explicitly allows. `get_code_service()` catches only `ValueError` and `CodeGenerationError`, so the request dies as an opaque 500 instead of a 4xx naming the unsupported problem type.

## Steps to Reproduce
1. Save a model but never set its problem type, so `model_basic.model_type` stays `NULL`.
2. Request the generated training script for that model (`GET /model/code/{model_name}`).
3. `_map_template()` raises `KeyError: None` and the response is a 500 with no usable message.

## Expected vs Actual
- **Expected:** A 4xx response whose message names the unsupported `model_type` and lists the supported problem types.
- **Actual:** An unhandled `KeyError` propagates out of `_map_template()` and surfaces as a 500.

## Acceptance Criteria
- [ ] `_map_template()` raises a descriptive error (`ValueError` or `CodeGenerationError`) naming the offending `model_type`, instead of letting `KeyError` escape.
- [ ] `model_type` of `None` is handled explicitly, not only unrecognised ints.
- [ ] `get_code_service()` maps the new error to a 4xx response rather than letting it become a 500.
- [ ] Regression test asserting `_map_template(None)` and `_map_template(99)` raise the descriptive error.
- [ ] The three existing `TestMapTemplate` cases for the valid `ProblemType` values still pass unchanged.

## Resources
- `tensormap-backend/app/services/code_generation.py:78-85` - `_map_template()`; line 85 is the unguarded `return options[problem_type_id]`
- `tensormap-backend/app/services/code_generation.py:37-42` - the null checks that already landed, for the error style to match
- `tensormap-backend/app/models/ml.py:26` - `model_type: int | None`, so `None` really does reach `_map_template()`
- `tensormap-backend/app/services/deep_learning.py:418-430` - `get_code_service()`, which catches `ValueError` (400) and `CodeGenerationError` (404) and nothing else
- `tensormap-backend/app/shared/enums.py:4-9` - `ProblemType` with values 1, 2, 3
- `tensormap-backend/tests/test_code_generation_service.py:73-84` - `TestMapTemplate`, where the regression test belongs

<details>
<summary>Original report by @VishakBaddur (2026-03-05)</summary>

## Describe the bug
`app/services/code_generation.py` has two unhandled crash scenarios:
**1. Missing null checks (AttributeError)**
`generate_code()` and `_file_location()` perform DB lookups without 
null checks - if model or file is not found, crashes with AttributeError.

**2. Unknown problem_type_id (KeyError)**
`_map_template()` accesses a dict with no fallback - unknown problem types crash with an unhandled KeyError instead of a clear error message.

## Expected behavior
- Return clear ValueError when model or file not found in DB
- Raise ValueError with descriptive message for unknown problem types

## Environment
- Backend: FastAPI + SQLModel
- File: `app/services/code_generation.py`

</details>

VishakBaddur:
## Summary
Fixes #195
`app/services/code_generation.py` had three unhandled crash scenarios.

## Changes

**1. `generate_code()`: null checks for model and file lookups**
- `model_configs` was used without checking for None: raises `ValueError` with clear message if model not found
- `file` was used without checking for None: raises `ValueError` with clear message if file not found

**2. `_map_template()` — KeyError guard for unknown problem types**
- Dict access `options[problem_type_id]` had no fallback: now raises `ValueError` for unknown problem types instead of crashing with `KeyError`

**3. `_file_location()`: null check for file lookup**
- Same pattern as `_helper_generate_file_location` in `model_run.py` (fixed in #194)


ivantha (commented):
## ◎ Argus Review

This PR correctly addresses three unhandled crash scenarios in `code_generation.py` by adding null checks and a `KeyError` guard, converting implicit crashes (`AttributeError`, `KeyError`) into explicit `ValueError` exceptions with descriptive messages. The changes are logically ordered and consistent with the fix pattern established in #194. However, using `ValueError` for "resource not found" scenarios in a web-service context may still surface as 500 Internal Server Errors to clients if no exception middleware exists, and the PR lacks test coverage for the new error paths. Overall this is a clear improvement in defensive coding, but a few follow-up concerns warrant attention.

VishakBaddur:
Hi @ivantha , addressed both findings:

- Added `model_configs.file_id is None` guard before the file lookup to fail fast with a clear error
- Refactored `_file_location` to accept the already-loaded `DataFile` object instead of re-querying the DB by ID: eliminates the redundant round trip

ivantha (commented):
## ◎ Argus Review

This PR correctly identifies and patches three real crash scenarios (missing model, missing file, and unknown problem type) and incidentally eliminates a pre-existing redundant database query. The error messages are clear and follow a consistent `ValueError` pattern. However, the most prominent gap is that `_file_location` — despite the PR description claiming it receives a null check — has no null guard in its own body, leaving future callers unprotected. Two additional issues worth addressing before merging: `file_type == None` produces a malformed path string (pre-existing but missed), and the PR description meaningfully mischaracterises the `_file_location` change, which is a signature refactor, not a null-check addition.

### ⚠️ Breaking Changes

- **[LOW]** Private `_file_location` function signature changed incompatibly: **Previous behavior**: `_file_location(file_id, db: Session) -> str` — accepted a raw integer ID and a database session; performed its own DB lookup internally.

**New behavior**: `_file_location(file: DataFile) -> str` — accepts a pre-fetched `DataFile` ORM object; no longer accesses the database.

**Who is affected**: Any code that calls `_file_location` directly with the old signature. Because the function is prefixed with `_` (Python convention for private/internal), external callers are unlikely, but the project's own tests or sibling modules may call it.

**Migration path**: Replace `_file_location(some_id, db)` with a pre-fetch + null check + `_file_location(file_obj)` pattern, matching the updated call site in `generate_code`.
  > **Suggestion:** If any tests directly call `_file_location`, update them to pass a `DataFile` mock instead of an integer ID and session mock. Search the test suite for `_file_location` references.
- **[INFO]** Exception type changed from `KeyError` / `AttributeError` to `ValueError` for all error paths: **Previous behavior**:
- `_map_template` with an unknown `problem_type_id` raised `KeyError`.
- Accessing `.file_id` on `None` raised `AttributeError`.
- Calling `_file_location` with a non-existent `file_id` resulted in `AttributeError` inside the function on `file.file_type`.

**New behavior**: All failure paths now raise `ValueError` with descriptive messages.

**Who is affected**: Any `try/except` blocks in callers that catch `KeyError` or `AttributeError` specifically will no longer intercept these failure cases.

**Migration path**: Callers should update `except KeyError` → `except ValueError` (or `except (KeyError, ValueError)` transiently). A broad `except Exception` catch is unaffected.


### PR Description Issues

- **[MEDIUM]** PR description misrepresents the `_file_location` change as a null check: Under **"3. `_file_location()`: null check for file lookup"**, the description says:

> *Same pattern as `_helper_generate_file_location` in `model_run.py` (fixed in #194)*

This implies a null check was added inside `_file_location`. In reality, the change is a **full signature refactoring**: the function no longer takes a `file_id` + `db` session — it now accepts a pre-loaded `DataFile` object and performs no DB access at all. No null check was added to the function body (as noted in finding #1 above, it is conspicuously absent).

This mismatch between the stated fix and the actual change could mislead reviewers into approving under the assumption that a null guard exists inside `_file_location` when it does not.
  > **Suggestion:** Replace the third bullet in the description with:

> **3. `_file_location()` — signature refactored to accept pre-loaded `DataFile`**
> - Removed the internal DB query from `_file_location`; the function now accepts an already-fetched `DataFile` object.
> - This eliminates a duplicate DB round-trip that existed in `generate_code` (the `file` variable was previously fetched and silently discarded).
> - Note: a null guard was **not** added to `_file_location` itself; the caller (`generate_code`) guards against `None` before the call.
- **[MEDIUM]** PR description has no testing section: The description has no mention of how the changes were validated — no unit tests added or updated, no manual testing steps, no regression checks. For a patch fixing crash scenarios, reviewers need confidence that:
1. The happy path still works end-to-end.
2. Each new `ValueError` is reachable and produces the expected message.
3. No existing tests were broken by the `_file_location` signature change.

Without this information, reviewers must assume the author tested manually and that nothing broke, which increases review risk.
  > **Suggestion:** Add a **Test plan** section to the description, for example:

```markdown
## Test plan
- [ ] Updated `test_generate_code` to mock a missing model and assert `ValueError` is raised
- [ ] Updated `test_generate_code` to mock a missing file and assert `ValueError` is raised  
- [ ] Updated `test_map_template` to pass an unknown problem type and assert `ValueError` is raised
- [ ] Updated `test_file_location` (if it exists) to pass a `DataFile` object instead of `(file_id, db)`
- [ ] Full test suite passes: `pytest tensormap-backend/tests/`
- [ ] Manual end-to-end: generated code for an existing model returns correct output
```
- **[LOW]** PR description does not mention the performance improvement (duplicate DB query eliminated): The refactoring of `_file_location` also incidentally fixes a pre-existing bug: `generate_code` queried `DataFile` but discarded the result, and then `_file_location` queried `DataFile` a second time. The PR eliminates this duplicate query. This is a meaningful side-effect worth surfacing to reviewers, both for credit and so they understand why the signature changed.
  > **Suggestion:** Add a note under change #3:

> Additionally, this removes a duplicate `DataFile` database query: `generate_code` previously fetched `file` but discarded it (dead code), while `_file_location` made a second identical fetch. Now the result of the first fetch is reused.

ivantha:

ivantha:
Thank you for the null checks in code_generation.py (#195)! This work has been superseded by #269 which addresses the same crash scenarios across code_generation, model_run, and data_upload in a unified fix. We appreciate the original bug investigation.

PR #198, opened 2026-03-06
fix: catch ValueError in get_code_service to return clean 400 response

### LINKED ISSUE #197: [Bug] get_code_service raises unhandled ValueError causing 500 instead of clean error response
## Describe the bug
`get_code_service()` in `app/services/deep_learning.py` calls `generate_code()` without a try/except block. Since PR #196 added 
`ValueError` raises to `generate_code()` for missing models/files, these exceptions now propagate uncaught to the router, returning an unhandled 500 instead of a clean 400 error response.

## Steps to Reproduce
1. Call `GET /api/v1/model/code/<model_name>` where the model exists in DB but its associated file has been deleted
2. `generate_code()` raises `ValueError`
3. Router returns 500 Internal Server Error instead of 400

## Expected behavior
`get_code_service()` should catch `ValueError` and return a clean 400 response with the error message.

## Suggested Fix
```python
def get_code_service(db, model_name, project_id=None):
    err = _verify_model_project(db, model_name, project_id)
    if err:
        return err
    try:
        python_code = generate_code(model_name, db)
    except ValueError as e:
        return _resp(400, False, str(e))
    return {"content": python_code, "file_name": model_name + ".py"}, 200
```

VishakBaddur:
## Summary
Fixes #197

## Change
`get_code_service()` called `generate_code()` without error handling. Since PR #196 added `ValueError` raises to `generate_code()` for missing models/files, these exceptions propagated uncaught to the router, returning 500 instead of a clean error response.

Added try/except to catch `ValueError` and return a proper 400 response with the error message.

### DIFF
--- a/tensormap-backend/app/services/deep_learning.py
+++ b/tensormap-backend/app/services/deep_learning.py
@@ -315,7 +315,17 @@ def get_code_service(db: Session, model_name: str, project_id: uuid_pkg.UUID | N
     err = _verify_model_project(db, model_name, project_id)
     if err:
         return err
-    python_code = generate_code(model_name, db)
+    try:
+        python_code = generate_code(model_name, db)
+    except ValueError as e:
+        logger.warning("Code generation failed for model '%s': %s", model_name, str(e))
+        return _resp(400, False, str(e))
+    except (FileNotFoundError, OSError):
+        logger.exception("File error during code generation for model '%s'", model_name)
+        return _resp(500, False, "Failed to read model file")
+    except Exception:
+        logger.exception("Unexpected error during code generation for model '%s'", model_name)
+        return _resp(500, False, "Internal error generating code")
     return {"content": python_code, "file_name": model_name + ".py"}, 200
 
 


ivantha (commented):
## ◎ Argus Review

This PR correctly addresses the uncaught `ValueError` propagation introduced by PR #196, preventing unintended 500 responses. The core fix is sound and appropriately scoped — the try block is minimal and the 400 status code is a reasonable choice for client-attributable input errors. However, there are several gaps: raw exception messages are forwarded directly to the client (information disclosure risk), exception handling is incomplete (non-`ValueError` exceptions from `generate_code()` are still unhandled), no tests were added, and there is no error logging.

ivantha (commented):
## ◎ Argus Review

This is a well-motivated, focused bug-fix PR that correctly addresses uncaught `ValueError` exceptions propagating to the router. The exception hierarchy is ordered correctly, logging levels are appropriately differentiated (`warning` vs `exception`), and 500 responses wisely use opaque messages to avoid information leakage. The main concerns are: (1) the `ValueError` handler returns `str(e)` verbatim to clients — the safety of those messages should be explicitly audited; (2) no tests accompany the change, leaving all three new branches untested; and (3) the PR description understates the scope by mentioning only `ValueError` while the diff introduces two additional handlers.

### ⚠️ Breaking Changes

- **[INFO]** Behavioral change: exceptions from `generate_code()` are now caught rather than propagated: **Previous behavior**: Any exception raised by `generate_code()` (including `ValueError`, `FileNotFoundError`, `OSError`, or any other `Exception` subclass) propagated uncaught through `get_code_service()` to the router, typically resulting in an unhandled 500 response.

**New behavior**: All exceptions from `generate_code()` are now caught inside `get_code_service()` and converted into structured `(dict, int)` response tuples — 400 for `ValueError`, 500 for everything else.

**Who is affected**: Any code that calls `get_code_service()` and relied on exceptions propagating (e.g. a caller with its own `except ValueError` wrapper, or a test that asserts an exception is raised). In practice this is unlikely given the function is a service layer intended to always return an HTTP tuple, but callers should be verified.

**Migration path**: If any caller wraps `get_code_service()` in a try/except, those handlers will no longer be reached for these exception types. Callers should instead inspect the returned HTTP status code.


### PR Description Issues

- **[LOW]** PR description understates scope — additional exception types not mentioned: The PR title reads *"fix: catch ValueError in get_code_service to return clean 400 response"* and the description's **Change** section says:

> *Added try/except to catch `ValueError` and return a proper 400 response with the error message.*

However, the actual diff adds handlers for **three** exception categories:
- `ValueError` → 400
- `FileNotFoundError, OSError` → 500
- `Exception` (catch-all) → 500

The broader `OSError` and catch-all blocks are not mentioned at all. Reviewers relying solely on the description may not scrutinise those branches, reducing the effectiveness of code review.
  > **Suggestion:** Update the **Change** section to reflect the full scope:

```
## Change
`get_code_service()` called `generate_code()` without error handling. Since PR #196 added
`ValueError` raises for missing models/files, these exceptions propagated uncaught, returning
500 instead of a clean error response.

Added structured try/except around the `generate_code()` call:
- `ValueError` → 400 with the error message (user-facing, expected failures)
- `FileNotFoundError / OSError` → 500 with generic message (infra-level failures)
- `Exception` (catch-all) → 500 with generic message (unexpected failures, logged with full traceback)
```
- **[LOW]** PR description is missing a testing section: The description explains *what* changed and *why*, but contains no information about how the fix was validated. There is no mention of:
- Manual testing steps performed
- Whether existing tests were run and passed
- Whether new tests were added (they were not, per the diff)

For a bug-fix PR that changes error-handling behaviour, reviewers should be able to verify the fix was exercised.
  > **Suggestion:** Add a **Testing** section:

```
## Testing
- [ ] Manually triggered the endpoint with a model name that causes `generate_code()` to raise
  `ValueError` and confirmed a 400 response is returned.
- [ ] Confirmed existing test suite passes (`pytest tensormap-backend/`).
- Note: unit tests for the new exception branches are tracked in issue #XXX.
```

ivantha:

ivantha:
Thank you for catching the ValueError in get_code_service (#197)! This fix was already landed in merged PR #226 which covers the identical change. We appreciate the bug report and the clean fix.

PR #203, opened 2026-03-10
fix: replace _VALID_TRANSFORMATIONS with _TRANSFORMATION_REGISTRY dispatch

### LINKED ISSUE #204: Bug: 4 of 7 data transformations rejected by backend validation
## Summary

The `_VALID_TRANSFORMATIONS` whitelist in `preprocess_data()` only contains 3 entries, but the function implements handlers for 7 transformations. The 4 missing ones are rejected with HTTP 422 before their handler code can execute.

The root cause is a design flaw: the validation set and the handler `if/elif` chain are two disconnected sources of truth with no enforcement to keep them in sync.

## Affected Transformations

| Transformation | Has Handler | In Whitelist | Status |
|---|---|---|---|
| One Hot Encoding | Yes | Yes | **Working** |
| Categorical to Numerical | Yes | Yes | **Working** |
| Drop Column | Yes | Yes | **Working** |
| Min-Max Normalization | Yes | **No** | **Broken** |
| Z-score Standardization | Yes | **No** | **Broken** |
| Log Transform | Yes | **No** | **Broken** |
| Fill Missing Values | Yes | **No** | **Broken** |

## Location

- **File:** `tensormap-backend/app/services/data_process.py`, line 233
- **Frontend UI:** `tensormap-frontend/src/components/Process/TransformationCreator.jsx`, lines 13–21

## Steps to Reproduce

1. Upload a CSV file via the frontend
2. Navigate to the data processing page
3. Select a column and choose "Min-Max Normalization" (or any of the 4 broken transforms)
4. Click apply
5. Backend returns: `{"success": false, "message": "Unknown transformation 'Min-Max Normalization'. Valid options: ['Categorical to Numerical', 'Drop Column', 'One Hot Encoding']"}`

## Root Cause

The 4 newer transformations were added to:
- The handler logic (lines 277–303)
- The frontend dropdown (TransformationCreator.jsx)

But the `_VALID_TRANSFORMATIONS` set was never updated. The validation set and handler `if/elif` chain are two disconnected sources of truth — every new transformation requires updating both, with no compiler or linter enforcement.

## Impact

- 57% of preprocessing features are broken (4 out of 7)
- Users see these options in the UI but get errors when using them
- The pattern will recur whenever a new transformation is added

## Proposed Fix

Replace `_VALID_TRANSFORMATIONS` set + `if/elif` chain with a single `_TRANSFORMATION_REGISTRY` dispatch dict. This makes it structurally impossible to add a handler without registering it, eliminating the class of bug entirely.

kartikeyg0104:
## What

Replaces the disconnected `_VALID_TRANSFORMATIONS` set + `if/elif` chain with a single `_TRANSFORMATION_REGISTRY` dict that maps transformation names to handler functions. Adds parametrized validation tests.

## Why

4 of 7 data transformations (Min-Max Normalization, Z-score Standardization, Log Transform, Fill Missing Values) were rejected with HTTP 422 despite having working handler code. The root cause was two disconnected sources of truth: a validation set and an `if/elif` chain that could silently drift apart.

## Before → After

```python
# BEFORE: Two disconnected sources of truth
_VALID_TRANSFORMATIONS = {"One Hot Encoding", "Categorical to Numerical", "Drop Column"}
# ... handler if/elif chain 25 lines below (easily drifts)

# AFTER: Single source of truth
_TRANSFORMATION_REGISTRY = {
    "One Hot Encoding": _one_hot_encode,
    "Categorical to Numerical": _categorical_to_numerical,
    "Drop Column": _drop_column,
    "Min-Max Normalization": _min_max_normalize,
    "Z-score Standardization": _zscore_standardize,
    "Log Transform": _log_transform,
    "Fill Missing Values": _fill_missing,
}
# Validation: `if name not in _TRANSFORMATION_REGISTRY`
# Dispatch:   `_TRANSFORMATION_REGISTRY[name](df, col, params)`
```

## Safety

- All 7 handler functions preserve identical behavior - extracted verbatim from the original `if/elif` branches
- Backward compatible - same transformation names, same API contract
- Matches frontend UI list in `TransformationCreator.jsx`
- Unknown names still rejected with 422
- Structurally prevents this class of bug from recurring
- Ruff lint + format passing

Fixes #204


ivantha (commented):
## ◎ Argus Review

This PR makes a correct and clean architectural fix: replacing a disconnected validation set + `if/elif` dispatch with a unified `_TRANSFORMATION_REGISTRY` dict, directly resolving the root cause of the 422 regression for 4 transformations. The extracted handlers are faithful to the original logic and the registry pattern is a textbook improvement. However, three data-integrity issues stand out: `_zscore_standardize` uses exact float equality (`std == 0`) while the sibling function `_min_max_normalize` correctly uses `np.isclose`; `_log_transform` silently skips the entire column transformation and still returns HTTP 200 when any value is below -1; and `_categorical_to_numerical` silently maps `NaN` to `-1` via `pd.Categorical.codes`, which can corrupt downstream ML models. The new parametrized validation tests are a welcome addition but only verify acceptance/rejection, not transformation correctness.

ivantha:

kartikeyg0104:
> This PR is behind the base branch. Please rebase onto the latest `main` so we can re-run CI and merge.

Updated the branch and squash it into a single commit.

kartikeyg0104:
@ivantha @rajutkarsh07 Please have a look

PR #221, opened 2026-03-13
fix(backend): return clear error when target field is missing

### LINKED ISSUE #212: [Bug] Model training crashes with KeyError if target_field not set in training config
## Describe the bug

When training a tabular classification model, if the `target_field` is not set in the training configuration, the backend crashes with a cryptic pandas `KeyError` instead of a clean validation error.

## Location

**File:** `tensormap-backend/app/services/model_run.py`  
**Lines:** 165-166

```python
def _run(model_name: str, db: Session) -> None:
    model_configs = db.exec(select(ModelBasic)...
    # ...
    else:  # Tabular problem (classification/regression)
        file_location = _helper_generate_file_location(db, file_id=model_configs.file_id)
        features = pd.read_csv(file_location)
        features.dropna(inplace=True)
        features = features.sample(frac=1, random_state=42).reset_index(drop=True)
        
        X = features.drop(model_configs.target_field, axis=1)  # ← CRASH if target_field is None
        y = features[model_configs.target_field]  # ← CRASH if target_field is None
```

## Crash output

```
KeyError: None
```

Instead of a user-friendly message.

## Root cause

No validation that `model_configs.target_field` is set before using it in pandas operations. If the user forgets to set the target field in the training configuration, they get a cryptic error.

## Expected behavior

The handler should validate that `target_field` is set before attempting to use it. If not set, raise a clear error:

```
"Training configuration incomplete: target field not set. Please set the target field in training configuration."
```

## Suggested fix

Add validation at the start of the tabular branch:

```python
if model_configs.target_field is None:
    raise ValueError("Training configuration incomplete: target field is required for tabular models")
```

## Impact

- **Severity:** High (training crashes)
- **Likelihood:** Medium (depends on UI preventing incomplete configs)
- **User experience:** Cryptic error message instead of helpful guidance

shivv23:
## Summary
- validate 	arget_field before tabular model training starts
- raise a clear ValueError instead of allowing a pandas KeyError: None`n- add a unit test that verifies no file lookup happens when target field is missing

Fixes #212


ivantha (commented):
## ◎ Argus Review

This is a clean, focused bug-fix PR that correctly addresses issue #212 by adding an early guard for a missing `target_field` before tabular model training begins. The production change is minimal and well-placed; the accompanying test is solid in structure and includes a valuable `assert_not_called()` assertion. The main concerns are: (1) a whitespace-only `target_field` silently bypasses the current falsy check and would still produce an obscure downstream error; (2) the parenthesised `with` syntax in the test requires Python 3.10+ and could silently break on older environments; and (3) the PR description contains a typo and lacks motivation/context and a test plan section. The breaking-change surface is very low — callers catching `KeyError` specifically would be affected, but that scenario was already an unhandled crash.

### ⚠️ Breaking Changes

- **[INFO]** Behavioral change: missing target_field now raises ValueError instead of KeyError: **Previous behavior**: When `target_field` was `None` on a tabular model config, `_run` would proceed past the validation point, load the CSV via `pd.read_csv`, call `dropna`, and eventually raise a `KeyError: None` from pandas when attempting to index a column by the `None` key.

**New behavior**: A `ValueError` with a descriptive message is raised immediately, before any file I/O.

**Who is affected**: Any caller or middleware that catches `KeyError` specifically (rather than a base `Exception`) to handle training failures will no longer intercept this error path. The socket.io error handler and any HTTP error boundary must handle `ValueError`.

**Migration path**: Audit any `except KeyError` blocks in the call stack above `_run` and ensure `ValueError` is also handled, or broaden the catch to `except (KeyError, ValueError)` / `except Exception`. Given this was previously an unhandled crash scenario, the practical impact is likely minimal.
  > **Suggestion:** Verify the top-level error handler for `_run` (e.g., the socket.io task wrapper) catches `ValueError` and routes it to the same user-facing error emission as other failures.


### PR Description Issues

- **[LOW]** Typo in PR description body: The first bullet point reads `"validate 	arget_field"` — there is a stray tab character and the leading `t` of `target_field` is missing. While minor, it undermines the professionalism of the description and could confuse readers skimming quickly.
  > **Suggestion:** Correct to:
```
- validate `target_field` before tabular model training starts
```
- **[MEDIUM]** PR description lacks motivation and user-impact context: The description lists *what* was changed but omits *why* it matters:
- What user-facing symptom did issue #212 report? (e.g., "training job hangs", "500 error with no message", "confusing stack trace in the UI")
- What did the raw `KeyError: None` look like to the end user or in server logs?
- Is this a regression, a long-standing paper-cut, or a newly discovered edge case?

Without this context, reviewers cannot assess severity or prioritise the fix relative to other open issues.
  > **Suggestion:** Expand the Summary section, for example:

```markdown
## Summary
Fixes #212. When a tabular model is submitted for training without a `target_field` set, the
training service would proceed to load the CSV and then crash with an unguarded `KeyError: None`
from pandas. This error surfaced as a generic 500 / socket.io error event with no actionable
message for the user.

This PR adds an explicit guard at the start of the tabular training path that raises a
`ValueError` with a human-readable message before any file I/O occurs, and adds a unit test
confirming that the file-location helper is never reached in this scenario.
```
- **[LOW]** No testing or verification section in PR description: The description does not include a testing plan or steps to reproduce and verify the fix. Reviewers and QA cannot easily validate the change manually, and there is no indication of whether the existing test suite was run against this branch.
  > **Suggestion:** Add a **Test plan** section:

```markdown
## Test plan
- [ ] `pytest tensormap-backend/tests/test_model_run.py` — all tests pass, including the new `test_raises_clear_error_when_target_field_missing`
- [ ] Manually submit a model training job with no target field set via the UI and confirm the returned error message is human-readable
- [ ] Confirm existing training jobs with a valid target field are unaffected
```

shivv23:

ivantha:
Solid defensive fix — adds an early `target_field` validation guard before the file lookup so users get a descriptive `ValueError` instead of a cryptic pandas `KeyError: None`. The test correctly asserts that the file-location helper is never called when validation fails. All CI passes.

ivantha:
The branch is behind main and needs a rebase. Please run `git fetch origin && git rebase origin/main` on the `codex/fix-212-target-field-validation` branch, resolve any conflicts, and force-push to re-trigger CI.

shivv23:

ivantha:
Thank you for the target_field validation fix! This work is covered by #255 which addresses the same missing-target-field error path more surgically. We appreciate the unit test and the systematic approach to these validation gaps.

PR #222, opened 2026-03-13
fix(backend): fail gracefully when model config is missing in runner

shivv23:
## Summary
- add an explicit null-check after loading ModelBasic in _run`n- raise a clear ValueError when a model name is not found in DB
- add a unit test to ensure missing model config is handled before file lookup

Part of #193


ivantha (commented):
## ◎ Argus Review

This PR adds a necessary null-guard in `_run` to prevent an opaque `AttributeError` when a model config is missing, which is a clear improvement. However, the primary concern is that the new test exercises the **public** `model_run` function rather than the private `_run` where the fix lives — given the class is named `TestModelRunEmitsErrorOnFailure` and a sibling test verifies socket.io error emission, there is a strong likelihood that `model_run` catches exceptions and emits them rather than re-raising, which would make `pytest.raises(ValueError)` either wrong or misleading. The fix itself is sound, but the test's correctness and the propagation of the `ValueError` through `model_run` need explicit verification before merging.

### ⚠️ Breaking Changes

- **[LOW]** Exception type changed from implicit `AttributeError` to explicit `ValueError`: **Previous behavior**: When `model_configs` was `None`, the next line `model_configs.model_type` would raise `AttributeError: 'NoneType' object has no attribute 'model_type'`.

**New behavior**: A `ValueError` is now raised with the message `"Model '{name}' not found in database"` before any attribute access.

**Who is affected**: Any caller of `_run` (or `model_run` if it re-raises) that explicitly catches `AttributeError` to handle the missing-config case would silently stop catching the error. In practice this is unlikely since `AttributeError` is a bad thing to catch here, but it is a behavioral change.

**Migration path**: Update any `except AttributeError` guards around calls to `_run` / `model_run` to `except (ValueError, ModelNotFoundError)` instead.
  > **Suggestion:** No immediate action needed since catching `AttributeError` for this purpose would have been a code smell anyway. Document the change in the PR description and ensure downstream error handlers check for `ValueError`.


### PR Description Issues

- **[LOW]** Malformed markdown in PR description body: The description contains a formatting artifact: `` _run`\n `` — the backtick and literal `\n` appear to be a broken inline code span with an accidental escaped newline. This suggests the description was written in a tool that mishandled newlines inside a list item.

**Original text**: `add an explicit null-check after loading ModelBasic in _run\`n`

This makes the description harder to read and signals a lack of polish.
  > **Suggestion:** Correct the formatting:

```
## Summary
- Add an explicit null-check after loading `ModelBasic` in `_run`
- Raise a clear `ValueError` when a model name is not found in the DB
- Add a unit test to ensure missing model config is handled before file lookup

Fixes #193
```
- **[LOW]** PR description missing motivation/context and previous failure mode: The description lists *what* was changed but not *why* or *what broke before*. A reviewer or future archeologist has no way to understand:
- What was the observable failure in #193? (crash? silent wrong result? 500 error?)
- What did the old exception look like to the end user? (`AttributeError: 'NoneType' object...`)
- Why is `ValueError` the right exception type here?
- Is there any user-visible change (e.g., the error message surfaced to the UI changed)?
  > **Suggestion:** Add a **Motivation** section:

```markdown
## Motivation
Previously, if a model name was passed to `_run` that did not exist in the
database, `db.exec(...).first()` returned `None` and the next line
(`model_configs.model_type`) raised an opaque `AttributeError`. This
caused confusing stack traces and made the root cause hard to diagnose.
This PR adds an explicit guard that raises a descriptive `ValueError` early.
```
- **[INFO]** PR description does not mention how the fix was tested end-to-end: The description mentions a unit test was added but does not describe any manual testing, integration testing, or how the fix was validated against the real database or a running instance. For a bug fix related to a live runner service, reviewers would benefit from knowing whether the fix was tested against an actual missing-model scenario.
  > **Suggestion:** Add a **Test Plan** section:

```markdown
## Test Plan
- [x] New unit test `test_raises_clear_error_when_model_config_missing` added and passes
- [ ] Manually triggered a run with a non-existent model name and confirmed the
      error message appears correctly in logs / socket.io response
- [ ] Existing test suite passes (`pytest tensormap-backend/tests/`)
```

shivv23:

ivantha:
Good defensive fix — adds a null-check after the `ModelBasic` DB query so a missing model name raises a clear `ValueError` rather than propagating an `AttributeError` on `None`. The test correctly verifies that the file-location helper is never reached. All CI passes.

ivantha:
The branch is behind main and needs a rebase. Please run `git fetch origin && git rebase origin/main` on the `codex/fix-193-missing-model-config` branch, resolve any conflicts, and force-push to re-trigger CI.

shivv23:

ivantha:
Thank you for the graceful model config null-check! This fix is covered by #269 which addresses the same missing-record scenarios across multiple backend services. We appreciate the detailed unit test included here.

PR #223, opened 2026-03-13
fix(backend): guard missing file record in model_run helper

shivv23:
## Summary
- add null-check in _helper_generate_file_location when DB file lookup returns none
- raise a clear ValueError instead of crashing with AttributeError on ile.file_type`n- add a unit test for the missing-file path

Part of #193

### DIFF
--- a/tensormap-backend/app/services/model_run.py
+++ b/tensormap-backend/app/services/model_run.py
@@ -88,6 +88,8 @@ def _helper_generate_file_location(db: Session, file_id) -> str:
     """Resolve the on-disk path for a dataset file by its DB ID."""
     upload_folder = get_settings().upload_folder
     file = db.exec(select(DataFile).where(DataFile.id == file_id)).first()
+    if file is None:
+        raise ValueError(f"File with id {file_id} not found in database")
     if file.file_type == "zip":
         return upload_folder + "/" + file.file_name
     return upload_folder + "/" + file.file_name + "." + file.file_type
--- a/tensormap-backend/tests/test_model_run.py
+++ b/tensormap-backend/tests/test_model_run.py
@@ -3,7 +3,7 @@ from unittest.mock import MagicMock, patch
 import pytest
 
 from app.services.deep_learning import get_code_service, run_code_service
-from app.services.model_run import model_run
+from app.services.model_run import _helper_generate_file_location, model_run
 from app.shared.enums import ProblemType
 
 
@@ -24,6 +24,14 @@ def _make_db(model_config: MagicMock) -> MagicMock:
     db = MagicMock()
     db.exec.return_value.first.return_value = model_config
     return db
+
+
+def test_helper_generate_file_location_raises_when_file_missing():
+    db = MagicMock()
+    db.exec.return_value.first.return_value = None
+
+    with pytest.raises(ValueError, match="not found in database"):
+        _helper_generate_file_location(db, "missing-file-id")
 
 
 class TestModelRunEmitsErrorOnFailure:


ivantha (commented):
## ◎ Argus Review

This is a focused, well-scoped bug-fix PR that correctly replaces an unguarded `AttributeError` crash with an explicit `ValueError`, making the failure mode intentional and diagnosable. The core guard logic and error message are sound, and the test correctly uses `pytest.raises` with a `match` assertion. However, the guard is placed *after* an unconditional `get_settings()` call, which means the test is implicitly environment-dependent (and slightly inefficient in the error path) — moving the guard before `get_settings()` would be the cleanest resolution. The test suite would also benefit from covering the success branches and mocking `get_settings()` explicitly. No public APIs or signatures are changed, so this is a patch-level change.

### ⚠️ Breaking Changes

- **[INFO]** Exception type changed from implicit `AttributeError` to explicit `ValueError`: **Previous behavior**: When `file_id` resolved to no DB row, `db.exec(...).first()` returned `None` and the next statement (`file.file_type`) raised `AttributeError: 'NoneType' object has no attribute 'file_type'`.

**New behavior**: The guard raises `ValueError(f"File with id {file_id} not found in database")`.

**Who is affected**: Any caller of `_helper_generate_file_location` (or the public functions that delegate to it, such as `model_run`) that catches `AttributeError` to detect the missing-file condition — though this pattern would be highly unusual and incorrect to rely on.

**Migration path**: No migration is expected to be necessary; the previous `AttributeError` was an unintentional crash, not a designed contract. However, if any `except AttributeError` blocks exist in calling code to handle this case, they must be updated to `except ValueError`.


### PR Description Issues

- **[LOW]** Typographical / markdown rendering error in PR description: The description contains a garbled fragment: `"ile.file_type\`n"`. This appears to be a malformed inline code span — likely intended to read `file.file_type` followed by a newline — caused by a missing backtick or incorrect escape. It makes the description harder to parse.
  > **Suggestion:** Replace the garbled text with the intended sentence, e.g.:

> raise a clear `ValueError` instead of crashing with `AttributeError` on `file.file_type`
- **[LOW]** PR description lacks motivation / user-impact context: The description explains *what* changed (null-check, ValueError, unit test) but not *why* it matters: What triggered issue #193? What was the user-visible impact (a 500 error? silent data corruption? a crash in a background worker)? Without this context, reviewers cannot assess severity or decide whether the fix is complete.
  > **Suggestion:** Add a **Motivation** section, for example:

```
## Motivation
When a `model_run` job references a `file_id` that no longer exists in the DB
(e.g. after a file is deleted while a run is queued), the helper crashed with an
unhandled `AttributeError`, producing a confusing 500 response with no
actionable message. This fix surfaces the problem as a clear `ValueError` so
the error can be caught and returned as a 404/422 at the API boundary.
```
- **[LOW]** PR description does not describe how the fix was manually verified end-to-end: The description mentions a unit test but provides no information about whether the fix was validated in a running environment (e.g. by triggering a `model_run` with a deleted file and observing the improved error response). For a bug fix affecting runtime error handling, reviewers typically want confidence that the happy path is also unaffected.
  > **Suggestion:** Add a **Test plan** section:

```
## Test plan
- [ ] `pytest tensormap-backend/tests/test_model_run.py` passes (new + existing tests)
- [ ] Triggered a `model_run` referencing a deleted file; confirmed API returns a
      structured error instead of a 500 `AttributeError` traceback
- [ ] Triggered a `model_run` with a valid file; confirmed behaviour is unchanged
```

shivv23:

ivantha:
Good defensive fix — guards the `_helper_generate_file_location` helper against a missing `DataFile` DB record, raising a descriptive `ValueError` instead of an `AttributeError` on `None`. Test correctly exercises the missing-file path. All CI passes.

ivantha:
The branch is behind main and needs a rebase. Please run `git fetch origin && git rebase origin/main` on the `codex/fix-193-missing-file-record` branch, resolve any conflicts, and force-push to re-trigger CI.

shivv23:

ivantha:
Thank you for the null-check guard in model_run! This fix has been superseded by #269 which addresses the same unhandled error patterns across model_run, code_generation, and data_upload services in a single focused pass. We appreciate the systematic error-handling work.

PR #226, opened 2026-03-13
fix(backend): catch code generation ValueError in get_code_service

### LINKED ISSUE #197: [Bug] get_code_service raises unhandled ValueError causing 500 instead of clean error response

shivv23:
## Summary
- wrap generate_code() in get_code_service with ValueError handling
- return standard 400 response with the error message instead of bubbling to 500
- add unit tests for status code and message contract

Fixes #197

### DIFF
--- a/tensormap-backend/tests/test_model_run.py
+++ b/tensormap-backend/tests/test_model_run.py
@@ -2,7 +2,7 @@ from unittest.mock import MagicMock, patch
 
 import pytest
 
-from app.services.deep_learning import run_code_service
+from app.services.deep_learning import get_code_service, run_code_service
 from app.services.model_run import model_run
 from app.shared.enums import ProblemType
 
@@ -84,3 +84,20 @@ class TestRunCodeServiceOnFailure:
             result, _ = run_code_service(db, "my_model")
 
         assert result["success"] is False
+
+
+class TestGetCodeServiceOnFailure:
+    def test_returns_http_400_when_generate_code_raises_value_error(self):
+        db = MagicMock()
+        with patch("app.services.deep_learning.generate_code", side_effect=ValueError("file missing")):
+            _, status_code = get_code_service(db, "my_model")
+
+        assert status_code == 400
+
+    def test_response_body_has_error_message_from_value_error(self):
+        db = MagicMock()
+        with patch("app.services.deep_learning.generate_code", side_effect=ValueError("file missing")):
+            result, _ = get_code_service(db, "my_model")
+
+        assert result["success"] is False
+        assert result["message"] == "file missing"
--- a/tensormap-backend/app/services/deep_learning.py
+++ b/tensormap-backend/app/services/deep_learning.py
@@ -316,7 +316,10 @@ def get_code_service(db: Session, model_name: str, project_id: uuid_pkg.UUID | N
     err = _verify_model_project(db, model_name, project_id)
     if err:
         return err
-    python_code = generate_code(model_name, db)
+    try:
+        python_code = generate_code(model_name, db)
+    except ValueError as e:
+        return _resp(400, False, str(e))
     return {"content": python_code, "file_name": model_name + ".py"}, 200
 
 


tensormap-backend/tests/test_model_run.py
```diff
@@ -84,3 +84,20 @@ def test_response_body_has_success_false(self):
...
+
+        assert status_code == 400
+
+    def test_response_body_has_error_message_from_value_error(self):
+        db = MagicMock()
+        with patch("app.services.deep_learning.generate_code", side_effect=ValueError("file missing")):
+            result, _ = get_code_service(db, "my_model")
+
+        assert result["success"] is False
+        assert result["message"] == "file missing"
```
ivantha: **[LOW]** `get_code_service` tests placed in `test_model_run.py` — semantic mismatch

`test_model_run.py` by name concerns `run_code_service` and the `model_run` function (as evidenced by the existing imports). Embedding `TestGetCodeServiceOnFailure` here conflates two distinct service concerns and makes the test suite harder to navigate.

Future developers searching for `get_code_service` test coverage will not intuitively look in a file named `test_model_run.py`.

**Suggestion:**
```
Move `TestGetCodeServiceOnFailure` to a dedicated file, e.g. `test_get_code_service.py` or the more general `test_deep_learning.py` if one exists:

```
tensormap-backend/tests/test_get_code_service.py
```

This also removes the need to add `get_code_service` to the import line of `test_model_run.py`.
```

ivantha (commented):
## ◎ Argus Review

This is a focused, well-scoped bug-fix PR that correctly converts an unhandled `ValueError` from `generate_code()` into a client-friendly HTTP 400 response. The core implementation change is clean and uses an appropriately narrow exception type. However, the accompanying tests have a significant structural flaw: they never mock `_verify_model_project`, which means the actual `generate_code` error path may never be exercised if the verification step returns early — the tests could be silently passing against the wrong code path. Additionally, the raw exception message is forwarded directly to the client, which risks leaking internal file paths or implementation details. The PR description is functional but thin on context about the root cause.

### ⚠️ Breaking Changes

- **[INFO]** Behavioral change: `ValueError` from `generate_code` now yields HTTP 400 instead of propagating as 500: **Previous behavior**: An unhandled `ValueError` raised inside `generate_code()` would propagate up through `get_code_service`, reach the WSGI/ASGI error handler, and be returned as an HTTP 500 Internal Server Error (or re-raised as an unhandled exception in tests).

**New behavior**: The same `ValueError` is caught and converted to an HTTP 400 Bad Request response with the error message in the body.

**Who is affected**: Any API client or integration test that explicitly expected a 500 (or an exception) for this code path. Also affects error-monitoring dashboards that alert on uncaught exceptions — this error will no longer appear as an exception trace.

**Migration path**: No migration required for well-behaved clients that treat 4xx and 5xx distinctly. Monitoring configurations that key off uncaught exceptions should add an alert for 400 responses on the code-generation endpoint.
  > **Suggestion:** Ensure server-side error logging is still present for these 400 errors so operational observability is not lost:

```python
except ValueError as e:
    logger.warning("[get_code_service] ValueError for model '%s': %s", model_name, e)
    return _resp(400, False, str(e))
```


### PR Description Issues

- **[LOW]** PR description lacks motivation and context for when `generate_code` raises `ValueError`: The summary bullets describe *what* was changed but not *why* `generate_code` raises `ValueError` or under what user-facing conditions issue #197 is triggered. A reader reviewing this PR cold cannot tell:
- What user action causes the `ValueError`?
- Was `generate_code` always capable of raising it, or was this a regression?
- Are there other call sites of `generate_code` that need the same fix?

This context is important for understanding whether the fix is complete.
  > **Suggestion:** Expand the description with a **Motivation** section, for example:

```markdown
## Motivation
When a user requests code generation for a model whose associated file has been deleted or was never
uploaded, `generate_code()` raises a `ValueError` with the message `"file missing"`. Previously this
propagated as an unhandled exception, causing the server to return HTTP 500. Users saw an opaque
server error instead of an actionable message.

## What changed
- `get_code_service` now wraps `generate_code()` in a `try/except ValueError` block
- Returns HTTP 400 with the ValueError message so the client can surface it
- Unit tests added for both the status code and the response body contract

## Testing
- New `TestGetCodeServiceOnFailure` class with two tests (status code + message body)
- Manually reproduced #197 before and after the fix
```

shivv23:
@ivantha Thanks for the review! This PR is worth merging now

ivantha:
Correct error-boundary fix — wraps `generate_code()` in a `ValueError` handler inside `get_code_service` so domain validation errors surface as HTTP 400 with the error message rather than propagating as 500s. Tests verify both the status code and the response body contract. All CI passes.

ivantha:
The branch is behind main and needs a rebase. Please run `git fetch origin && git rebase origin/main` on the `codex/fix-197-get-code-error-handling` branch, resolve any conflicts, and force-push to re-trigger CI.

shivv23:

PR #227, opened 2026-03-13
Paginate dataset preview to avoid loading full CSVs

shivv23:
## Summary
- add `offset` and `limit` query params to `/data/process/file/{file_id}`
- return paginated payloads (`rows`, `columns`, `pagination`) instead of a full JSON string
- cache CSV metadata while serving preview data to reduce repeated reads
- add Previous/Next pagination controls in the dataset preview table

## Why
Large CSV previews currently load the entire file at once, which can make the UI lag or freeze.

## Testing
- updated backend service tests for paginated file preview
- verified modified backend modules compile with `python -m compileall`

### DIFF
--- a/tensormap-backend/app/routers/data_process.py
+++ b/tensormap-backend/app/routers/data_process.py
@@ -83,14 +83,14 @@ async def get_correlation(file_id: uuid_pkg.UUID, db: Session = Depends(get_db))
 
 
 @router.get("/data/process/file/{file_id}")
-async def get_file(
+def get_file(
     file_id: uuid_pkg.UUID,
-    page: int = Query(1, ge=1),
-    page_size: int = Query(50, ge=1, le=1000),
+    offset: int = Query(0, ge=0),
+    limit: int = Query(100, ge=1, le=500),
     db: Session = Depends(get_db),
 ):
-    """Return a paginated slice of a CSV file's contents with pagination metadata."""
-    body, status_code = get_file_data(db, file_id=file_id, page=page, page_size=page_size)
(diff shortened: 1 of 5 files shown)


shivv23:
@ivantha This PR has been adjusted accordingly making it worth merging. Thanks for the review!

shivv23:
@ivantha kindly approve for the pending workflow. Thanks!

ivantha:
This PR has a few issues that need addressing before it can be merged:

1. **Broken function signature**: The router calls `get_file_data(db, file_id=file_id, offset=offset, limit=limit)` but the service function `get_file_data` still has `page` and `page_size` parameters in its signature (visible in the diff context). The internal logic still references `page_size` and `skip`/`total_pages` — these need to be updated to use `offset`/`limit`.

2. **Duplicated/dead code in `add_target_service`**: The file-type check, column retrieval, and target validation block appears **twice** in the new code. The second copy (lines after the first `existing_rows` early-return block) is unreachable dead code and should be removed.

3. **Test coverage gap**: Several existing tests for pagination edge cases (empty file, out-of-range page) were removed without equivalent replacement tests for the new offset/limit scheme (e.g. offset beyond total rows).

Please fix these issues and push an update.

shivv23:
Thanks for the review @ivantha! After checking, the pagination feature and validation improvements from this PR have already been merged into main through other PRs. The current main already has:
1. Backend pagination with page/page_size in get_file_data 
2. Frontend pagination with Previous/Next buttons 
3. Proper error sanitization 
Since the changes are already in main, I'll close this PR as it's now stale.

PR #229, opened 2026-03-13
Validate and upsert dataset target field assignments

shivv23:
## Summary
- validate target field against actual CSV columns before saving
- make target assignment idempotent (create or update)
- clean up duplicate target rows for legacy data during update path
- return clearer status/messages for unchanged targets and invalid targets

## Why
Target assignment should be safe to repeat and should fail early when the requested column does not exist.

## Testing
- added service-level tests for validation, update, no-op, and header backfill paths
- verified backend modules compile with `python -m compileall`


tensormap-backend/app/services/data_process.py
```diff
@@ -26,18 +26,52 @@ def _get_file_path(file: DataFile) -> str:
...
+                return _resp(200, True, "Target field already set for this file")
+            target_record.target = target
+            db.add(target_record)
+            db.commit()
+            logger.info("Target field updated to '%s' for file_id=%s", target, file_id)
+            return _resp(200, True, "Target field updated successfully")
 
         data_process = DataProcess(file_id=file_id, target=target)
         db.add(data_process)
         db.commit()
```
ivantha: **[LOW]** add_target_service has grown complex enough to warrant extraction of helper methods

The function now handles four distinct responsibilities in sequence:
1. File existence check
2. CSV type check
3. Column backfill + target validation
4. Upsert logic with duplicate cleanup

At ~45 lines, the function is approaching a complexity level where individual paths become hard to test and reason about independently. The duplicate-cleanup branch is particularly non-obvious inline.

**Suggestion:**
```
Consider extracting:
```python
def _resolve_columns(db, file) -> list[str]:
    """Return cached columns, backfilling from disk if needed."""
    ...

def _upsert_target(db, file_id, target) -> tuple:
    """Handle create-or-update with duplicate cleanup."""
    ...
```
This also makes each logical piece independently unit-testable.
```

File: tensormap-backend/app/services/data_process.py [hunk 2]
```diff
@@ -26,18 +26,52 @@ def _get_file_path(file: DataFile) -> str:
...
         logger.info("Target field '%s' added for file_id=%s", target, file_id)
         return _resp(201, True, "Target field added successfully")
     except Exception as e:
+        db.rollback()
         logger.exception("Error storing record: %s", str(e))
```
ivantha: **[INFO]** Positive: db.rollback() added in exception handler — good defensive pattern

Adding `db.rollback()` in the `except` block is an important correctness improvement. Without it, a partially-executed transaction (e.g., `db.add()` called but `db.commit()` not yet reached) would be left pending, potentially causing session-level issues in connection-pooled environments. This is a good defensive practice that should be considered for other service functions in the module.

**Suggestion:**
```
No change needed. Consider auditing other service functions in the same file for missing rollback calls in exception handlers, or centralizing rollback in a context manager / middleware layer.
```

ivantha (commented):
## ◎ Argus Review

This PR introduces useful and well-intentioned improvements — idempotent upsert semantics, column validation, and header caching — that address real correctness gaps. The code is generally readable, tests cover the happy paths well, and the addition of `db.rollback()` in the exception handler is a good defensive fix. However, there are a handful of concrete bugs and gaps: the no-op detection condition is incorrect when duplicate rows exist (producing a misleading "updated" message), the column backfill is committed before target validation completes (a partial-write concern), and a race condition allows concurrent requests to re-create the duplicate rows the cleanup logic is trying to eliminate. Test coverage for the duplicate-cleanup path is missing entirely, and one test uses a fragile single-call mock that will silently fail if code order changes. Clients that check for HTTP 201 strictly should be alerted to the behavioral status-code change.

**Suggested Version Bump:** `minor`

### ⚠️ Breaking Changes

- **[MEDIUM]** HTTP 201 → 200 for existing-record paths breaks strict status-code consumers: **Previous behavior**: `add_target_service` returned `HTTP 201 Created` for every successful call, including calls that would have created duplicate `DataProcess` rows.

**New behavior**: Returns `HTTP 200 OK` when an existing record is updated or the target is already set; `HTTP 201` is now reserved for the first-time insert path.

**Who is affected**: Any API client, integration test, or frontend that checks `response.status === 201` to confirm success. These will now see `200` and may incorrectly treat a successful update as a failure or enter an error-handling branch.

**Migration path**: Update all consumers to accept both `200` and `201` as success indicators, or normalize the response in a shared response-handling layer.
- **[LOW]** New 400 error for non-CSV files changes previously permissive behavior: **Previous behavior**: Non-CSV files would pass through the file-type check (no check existed) and attempt to create a `DataProcess` record.

**New behavior**: Non-CSV files immediately return `HTTP 400` with `'Only CSV files support target field assignment'`.

**Who is affected**: Any caller that previously (intentionally or accidentally) called `add_target_service` with a non-CSV `file_id`.

**Migration path**: Callers should add a file-type guard before calling the service, or update error-handling to accept `400` as a validation failure.
- **[LOW]** New 422 error for invalid column names replaces previously undefined behavior: **Previous behavior**: If `target` was not a real column, the `DataProcess` record was created anyway with an invalid target value, likely causing downstream errors at model-training time.

**New behavior**: Returns `HTTP 422 Unprocessable Entity` with a descriptive message listing available columns.

**Who is affected**: Callers that supplied invalid column names (possibly relying on the permissive behavior to store partial/speculative configurations).

**Migration path**: This is a correctness improvement. Callers should validate the target column client-side before calling the API, and handle `422` as an expected validation error.


### PR Description Issues

- **[LOW]** PR description omits breaking change notice for HTTP status code behavior change: The description accurately summarizes the code changes but does not flag the behavioral change in HTTP response codes as a potential breaking change for consumers. Specifically:

- **Before**: `add_target_service` always returned `201` on success.
- **After**: Returns `200` for updates/no-ops and `201` only for new inserts. Also introduces `422` (invalid column) and `400` (non-CSV file type) as new error codes.

Any client or test that asserts `status == 201` unconditionally will break silently on the update/no-op paths.
  > **Suggestion:** Add a **Breaking Changes** section to the PR description:
```
## Breaking Changes
- `add_target_service` now returns `200` (not `201`) when an existing target record is updated or is already set. Callers that check `status == 201` must be updated to accept `200` as well.
- New `422` error response when the target column does not exist in the CSV.
- New `400` error response when the file type is not `csv`.
```
- **[INFO]** PR description is clear and well-structured — good template adherence: The PR description covers Summary, Why, and Testing sections with concrete bullet points. The stated goals (idempotency, early validation, clearer messages) align directly with the code changes. No unfilled template checkboxes or placeholder text are present.
  > **Suggestion:** No changes needed beyond adding the missing breaking-changes section noted above.

shivv23:
@ivantha kindly approve for the pending workflow. Thanks!

ivantha:
This PR has two issues that need fixing before it can be merged:

1. **Duplicated/dead code in `add_target_service`**: The file-type check, column retrieval, target validation, and `existing_rows` handling block appears **twice** in the new code. After the first block returns (on every path — update, no-op, or new create), the second copy is entirely unreachable dead code. Please remove the duplicate block.

2. **No CI checks ran**: The status check rollup is empty, meaning CI did not trigger on this PR. Please ensure the branch is up to date and CI runs (lint + rebase check) before this can be merged.

shivv23:
Thanks @ivantha! Both issues addressed:
1. Duplicate code removed: Removed the dead code block from add_target_service
2. CI triggered: Branch rebased and CI is now running

Ready for re-review.

ivantha:
Closing as stale during repository cleanup. Thanks for the contribution. Feel free to reopen against the latest main if you'd like this revisited.

PR #232, opened 2026-03-14
fix: invalidate DB columns cache after preprocessing (#119)

### LINKED ISSUE #119: [Bug] Preprocessing leaves DataFile.columns stale in the file list
**Difficulty:** Easy
**Area:** backend

## Description

## Steps to Reproduce
3. `GET /data/upload/file`.

## Expected vs Actual

## Acceptance Criteria

## Resources

<details>

### Description

```python
````

### The Impact

### Proposed Solution

</details>

nguyenhungduong:
Fixes #119

## Problem
`preprocess_data()` overwrites the CSV on disk but never updates 
the `columns` JSON field cached in `DataFile` DB record, causing 
stale column data in the API response.

## Solution
Added `_update_file_columns_cache(db, file, file_path)` helper that:
- Reads only the header row via `pd.read_csv(file_path, nrows=0)`
- Updates `DataFile.columns` in DB and commits
- Called immediately after `df.to_csv()` in `preprocess_data()`

### DIFF
--- a/tensormap-backend/app/services/data_process.py
+++ b/tensormap-backend/app/services/data_process.py
@@ -116,9 +116,7 @@ def delete_one_target_by_id_service(db: Session, file_id: uuid_pkg.UUID) -> tupl
     if not file:
         return _resp(400, False, "File doesn't exist in DB")
 
-    target_record = db.exec(
-        select(DataProcess).where(DataProcess.file_id == file_id)
-    ).first()
+    target_record = db.exec(select(DataProcess).where(DataProcess.file_id == file_id)).first()
     if not target_record:
         return _resp(400, False, "Target field doesn't exist")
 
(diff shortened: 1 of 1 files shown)


ivantha:
The fix is clean and well-scoped — the helper reads only the header row (avoiding memory pressure on large files), handles exceptions gracefully without failing the whole preprocess operation, and CI is fully green. The branch just needs a rebase onto main before it can be merged. Please rebase and push.

nguyenhungduong:
> The fix is clean and well-scoped — the helper reads only the header row (avoiding memory pressure on large files), handles exceptions gracefully without failing the whole preprocess operation, and CI is fully green. The branch just needs a rebase onto main before it can be merged. Please rebase and push.

Rebased onto main and force-pushed. Ready for review.

ivantha:

PR #253, opened 2026-03-22
fix: remove accidentally committed file 'n' from repo root

Ushashi-Tech:
## Description
Removes the accidental file `n'` that was committed to the repo root directory. 
This file has no purpose and should not be tracked in version control.

Fixes #209

## Type of Change
- [x] Bug fix

## How Has This Been Tested?
- [x] Manual testing

Verified that the file `n'` existed in the repo root, deleted it, 
and confirmed it no longer appears in `git status` or `ls`.

## Screenshots (if applicable)
N/A — file deletion, no visual changes.

## Checklist
- [x] My code follows the project's style guidelines
- [x] I have performed a self-review
- [x] My changes generate no new warnings
- [x] Tests pass locally

PR #257, opened 2026-03-23
chore: remove accidental root file n'

shivv23:
Description:

This PR removes an accidentally committed stray file (n') from the repository root. The file appears to have been created by mistake during development and should not be part of the codebase.

Changes:

Removed the accidental file n' from the root directory
Why: Keeping the repository clean by removing unintended files that were accidentally committed.


ivantha:
This PR is a duplicate of #253 — both delete the same accidentally committed file `n'`. Once #253 is rebased and merged, this PR will no longer have anything to contribute and can be closed. The branch is also behind main.

PR #277, opened 2026-04-03
 feat: wire in rate limiter with HTTPException handler and env var disable

shivv23:
Description:
Wire rate limiter into main.py with HTTPException handler
Add rate_limit_enabled config option to disable via env var

Changes:
app/config.py: Add rate_limit_enabled: bool = True
app/main.py: Conditionally add rate limit middleware, add HTTPException handler for 429 responses
Checklist:

 ✅ Rate limit enforced on training endpoints
 ✅ 429 returns JSON: {"success": false, "message": "...", "data": null}
 ✅ Can disable via RATE_LIMIT_ENABLED=false

### DIFF
--- a/tensormap-backend/app/config.py
+++ b/tensormap-backend/app/config.py
@@ -15,6 +15,7 @@ class Settings(BaseSettings):
     max_content_length: int = 200 * 1024 * 1024
     api_base: str = "/api/v1"
     debug: bool = True
+    rate_limit_enabled: bool = True
 
     model_config = {"env_file": ".env"}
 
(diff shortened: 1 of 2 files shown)


ivantha:
The PR imports `RateLimitMiddleware` and `get_training_limiter` from `app.rate_limiter`, but the `tensormap-backend/app/rate_limiter.py` module is not included in the diff — it appears to be missing from the branch entirely. The application will fail to start with an `ImportError`. Please add the `rate_limiter.py` implementation (and any new dependencies in `pyproject.toml`/`uv.lock`) to the branch, then this can be re-reviewed.

shivv23:
You're right @ivantha - the rate_limiter.py implementation is missing from this PR. That's because the implementation was added in PR #288 (which is now mergeable). This PR #277 just wires the rate limiter into main.py, but it depends on #288 being merged first.I'll close this PR and re-create it after #288 is merged so it has the full implementation.

PR #311, opened 2026-04-24
fix: delete orphaned DB record when JSON file write fails

### LINKED ISSUE #97: Database Orphan Bug: Model name permanently locked if JSON file save fails (OSError)

ridham2890052-rgb:
Fixes #97. `db.rollback()` does not remove a record committed in a prior transaction step, causing model names to be permanently locked after an OSError. Extracted `_safe_model_write()` helper used by both service functions. Calls `db.rollback()` before `db.delete()` to clear dirty session state, wraps cleanup in nested try/except so a secondary DB failure can't crash the request, moves success log to after confirmed commit, and broadens exception scope to catch `TypeError`/`ValueError` from `json.dumps()`.

## Type of Change
- [x] Bug fix

## How Has This Been Tested?
- Manually simulated OSError during file write and confirmed orphaned DB record is removed and model name is released
- Verified `db.rollback()` is called before `db.delete()` to prevent `InvalidRequestError`

## Checklist

### DIFF
--- a/tensormap-backend/app/services/deep_learning.py
+++ b/tensormap-backend/app/services/deep_learning.py
@@ -177,16 +177,43 @@ def model_validate_service(db: Session, incoming: dict, project_id: uuid_pkg.UUI
             if error in str(e):
                 return _resp(400, False, errors.err_msgs[error])
         return _resp(400, False, "Model saving failed. Please recheck the model configs")
+    
+    def _safe_model_write(db: Session, model: ModelBasic, file_path: str, content: dict) -> Optional[str]:
+        try:
+            serialized = json.dumps(content)
+        except (TypeError, ValueError):
+            try:
+                db.rollback()
+                db.delete(model)
+                db.commit()
+            except Exception:
+                db.rollback()
+                logger.exception("Model serialization failed — DB orphan cleanup also failed")
+            else:
+                logger.error("Model serialization failed — orphaned DB record cleaned up")
+            return "Model validated but failed to serialize"
 
-    try:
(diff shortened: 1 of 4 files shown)


ivantha:
Closing as stale during repository cleanup. Thanks for the contribution. Feel free to reopen against the latest main if you'd like this revisited.

PR #344, opened 2026-05-22
ci: add frontend unit tests workflow

shivv23:
## Summary

The 45 frontend unit tests across 9 test files in `tensormap-frontend/src/` were never executed in CI. This PR adds a dedicated workflow that runs vitest on every PR to `main`.

## Changes

- Added `.github/workflows/frontend-tests.yml` — runs `npm run test:ci` on PRs to `main`
- Added `test:ci` script to `package.json` (`vitest run --reporter=default`) for CI-optimized output

## Validation

- All 45 frontend unit tests pass locally (9 test files)
- No changes to application code

## Files Changed

- `tensormap-frontend/package.json` — added `test:ci` script
- `.github/workflows/frontend-tests.yml` — new frontend test workflow

### DIFF
--- a/.github/workflows/frontend-tests.yml
+++ b/.github/workflows/frontend-tests.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: 22
+          node-version: "22.12"
           cache: npm
           cache-dependency-path: tensormap-frontend/package-lock.json
 
@@ -24,4 +24,4 @@ jobs:
 
       - name: Run tests
         working-directory: tensormap-frontend
-        run: npm run test:ci
+        run: npm test
--- a/.github/workflows/frontend-tests.yml
+++ b/.github/workflows/frontend-tests.yml
@@ -1,0 +1,27 @@
+name: Frontend Tests
+
+on:
+  pull_request:
+    branches: [main]
+
+jobs:
+  frontend-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
+          cache: npm
+          cache-dependency-path: tensormap-frontend/package-lock.json
+
+      - name: Install dependencies
+        working-directory: tensormap-frontend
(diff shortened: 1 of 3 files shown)


ivantha:
### Automated review summary

**Triage:** ci  |  **Recommendation:** ready-for-maintainer

**Findings**
- **[minor/quality]** `.github/workflows/frontend-tests.yml:15` New workflow pins node-version: 22, but the existing frontend job in .github/workflows/lint.yml pins node-version: 20. The two frontend CI jobs now run on different Node majors. (package.json engines requires >=22.12.0, so 22 is actually the more correct value and lint.yml is the stale one — but the inconsistency should be reconciled, ideally bumping lint.yml to match and pinning a concrete minor like 22.12 since bare '22' can resolve to a 22.x below the engines floor.)
- **[nit/quality]** `tensormap-frontend/package.json:35` The added test:ci script is 'vitest run --reporter=default', which differs from the existing 'test' script ('vitest run') only by an explicit --reporter=default — and 'default' is already vitest's default reporter. The flag is a no-op, so test:ci is effectively a duplicate of test; the workflow could just call 'npm test' (or 'npm run test') and drop the extra script.

PR #344 adds a genuinely missing CI workflow that runs the 9 pre-existing frontend vitest files on PRs to main; no such workflow exists on main today, and the PR touches no application code. Two minor cleanups: the workflow pins Node 22 while lint.yml's frontend job pins Node 20 (reconcile them, and prefer a concrete minor like 22.12 to satisfy the engines >=22.12.0 floor), and the new test:ci script only adds vitest's already-default --reporter=default, so it duplicates the existing test script and could be dropped in favor of 'npm test'. On GSoC scope: this does not write tests (it only runs existing ones), so it is not a full implementation of Phase 1, but because Phase 1 explicitly reserves 'test infrastructure', flag it for the assigned TensorMap student (Kartikey Gupta) to confirm before merging rather than landing it as foundation work."

shivv23:
I have addressed the points raised in the review:

**Nit: node-version 22 vs 20 discrepancy with lint.yml**
- Pinned both `frontend-tests.yml` and `lint.yml` to `"22.12"` to satisfy `package.json` engines `>=22.12.0`.

**Nit: test:ci script duplicates test**
- Removed the `test:ci` script. The workflow now runs `npm test` directly.

Issue #120, opened 2026-02-25
GSoC 2026 — TensorMap: Complete Neural Network Design Studio


ivantha:
## Title

Complete Neural Network Design Studio with Training Visualization, Model Export, Interpretability, and Hyperparameter Tuning

## Abstract

TensorMap is a visual neural network builder where users drag layer nodes onto a ReactFlow canvas, connect them, and train Keras models. The tool generates Python code and runs training with real-time Socket.IO progress updates. However, only 4 layer types exist (Input, Dense, Flatten, Conv2D), training output is unstructured text, there's no model export, and adding a new layer requires edits across 6+ files.

This project transforms TensorMap into a complete neural network design studio across six phases:

1. **Foundation** — test infrastructure (pytest + Jest) and a layer registry refactor that makes adding new layers data-driven instead of requiring 6-file edits
2. **Layer expansion** — 11 new layer types including pooling, regularization, and recurrent layers, plus a new sequence problem type
3. **Training visualization + callbacks** — structured Socket.IO metrics, real-time loss/accuracy charts via recharts, early stopping, learning rate scheduling, model checkpointing, validation split
4. **Model export + comparison** — SavedModel/ONNX/TFLite export formats and a comparison dashboard with overlaid training curves
5. **Model interpretability + post-training analysis** — confusion matrix, classification report, feature importance (permutation), and prediction explorer for understanding trained models
6. **Hyperparameter tuning engine** — grid search and random search over optimizer, learning rate, batch size, and epochs with real-time progress via Socket.IO

## Goals

- Make adding new layers a simple, data-driven process (one registry entry instead of 6-file edits)
- Expand from 4 to 15 layer types covering CNNs, RNNs, and regularization
- Replace unstructured text training output with structured metrics and real-time charts
- Support model export in industry-standard formats
- Enable comparing multiple training runs side by side
- Provide post-training interpretability tools (confusion matrix, feature importance, prediction explorer) to help users understand model behavior
- Automate hyperparameter tuning with grid and random search strategies, with real-time trial progress tracking

## Expected Outcomes

- Adding a new layer type requires only a registry entry (data-driven) instead of 6+ file edits
- 15 layer types available (up from 4), covering CNNs, RNNs, and regularization
- Training progress shown as real-time interactive charts with structured metrics
- Models can be exported in SavedModel, ONNX, and TFLite formats
- Multiple training runs can be visually compared in a dashboard
[...]

PR #345, opened 2026-05-22
feat: add model name check and count endpoints

shivv23:
## Summary

Add two utility endpoints for model management that improve the developer and user experience.

## Changes

- **`GET /model/check-name`** — Check if a model name is available before saving (scoped to project)
- **`GET /model/count`** — Get the total count of saved models (optional `project_id` filter)
- 6 unit tests covering available/unavailable name checks, project scoping, and count cases

## Validation

- 6 new unit tests added (168 total backend tests pass)

### DIFF
--- a/tensormap-backend/app/routers/deep_learning.py
+++ b/tensormap-backend/app/routers/deep_learning.py
@@ -9,9 +9,11 @@ from sqlmodel import Session
 from app.database import get_db
 from app.schemas.deep_learning import ModelNameRequest, ModelSaveRequest, ModelValidateRequest, TrainingConfigRequest
 from app.services.deep_learning import (
+    check_model_name_service,
     delete_model_service,
     get_available_model_list,
     get_code_service,
+    get_model_count_service,
     get_model_graph_service,
     get_training_history_service,
     model_save_service,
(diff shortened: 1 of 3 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `tensormap-backend/app/services/deep_learning.py:558` check_model_name_service scopes the uniqueness query by project_id when provided, but the actual save paths (model_save_service line 247 and model_validate_service line 153) check name uniqueness GLOBALLY with no project_id filter. So GET /model/check-name?model_name=X&project_id=P can report available:true while a subsequent save of X fails with 'Model name already used'. The check endpoint's answer does not match save-time behavior, defeating its purpose.
- **[minor/bug]** `tensormap-backend/app/services/deep_learning.py:560` check_model_name_service queries the raw model_name, but model_save_service applies _sanitize_model_name (strip + alnum/-/_ validation, line 243) before its uniqueness check. A name with surrounding whitespace or invalid characters will be reported as available here yet either fail validation or collide after sanitization at save time, so the availability answer is unreliable. Consider sanitizing (and surfacing invalid-name errors) the same way save does.
- **[nit/quality]** `tensormap-backend/app/services/deep_learning.py:566` Both new service functions construct raw response dicts instead of using the existing _resp(status_code, success, message, data) helper (defined line 45) that every other function in this module uses. The shape matches the {success, message, data} convention so behavior is correct, but using _resp keeps the file consistent.
- **[nit/test]** `tensormap-backend/tests/test_deep_learning_service.py:142` The two test_filters_by_project_id tests assert `"project_id" in str(call_stmt)` against a real SQLModel select() statement (mock_db is a bare MagicMock). The substring 'project_id' appears in the compiled SQL whenever the where-clause is present, so the assertion passes, but it is brittle and does not actually verify the filter value or that filtering changes results. A stronger test would assert the query yields a project-scoped result.

PR #345 adds two read-only model-management endpoints (GET /model/check-name and GET /model/count) plus six unit tests; it does not overlap the reserved GSoC scope in issue #120, so it is safe to consider on its merits. The main issue to fix before merge: check-name scopes uniqueness by project_id, but the actual save paths (deep_learning.py lines 247 and 153) enforce name uniqueness globally, so the endpoint can report a name as available that then fails to save. It also skips the _sanitize_model_name step that save applies, and the new functions bypass the module's _resp() helper. Ask the author to align check-name's scoping and sanitization with model_save_service, then it's a reasonable enhancement.

shivv23:

**Major: check_name scopes by project_id but save checks globally**
- Removed the `project_id` filter from the uniqueness query so `check_model_name_service` now checks globally, matching `model_save_service` behavior at line 247.

**Minor: sanitization mismatch**
- Added a `_sanitize_model_name()` call before the uniqueness check so invalid names or names with surrounding whitespace are rejected consistently with save-time behavior.

**Nit: raw response dicts instead of _resp() helper**
- Both `check_model_name_service` and `get_model_count_service` now use the existing `_resp()` helper.

**Nit: brittle project_id filter test**
- Replaced `test_filters_by_project_id` with `test_rejects_invalid_name` and `test_strips_whitespace`, which actually verify the sanitization logic.

PR #346, opened 2026-05-22
feat: add per-route error boundaries and PageLoader component

shivv23:
## Summary

Improve frontend resilience by adding per-route error boundaries, a reusable ErrorFallback UI with navigation, and a PageLoader spinner component.

## Changes

- **ErrorBoundary.jsx**: Enhanced with ErrorFallback (navigation + retry) and PageLoader (spinner) exports
- **App.jsx**: Each route is now individually wrapped in an ErrorBoundary — a crash in one page won't take down the entire app
- **ErrorBoundary.test.jsx**: 9 unit tests covering error rendering, retry, fallback states, and loading UI

## Validation

- All 54 frontend unit tests pass (10 test files)

### DIFF
--- a/tensormap-frontend/src/App.jsx
+++ b/tensormap-frontend/src/App.jsx
@@ -1,7 +1,7 @@
 import { Suspense, lazy } from "react";
 import { BrowserRouter, Route, Routes, Navigate } from "react-router-dom";
 import AppTopBar from "./components/AppTopBar";
-import ErrorBoundary from "./components/ErrorBoundary";
+import ErrorBoundary, { PageLoader } from "./components/ErrorBoundary";
 import WorkspaceLayout from "./components/Workspace/WorkspaceLayout";
 
 const ProjectsPage = lazy(() => import("./containers/ProjectsPage/ProjectsPage"));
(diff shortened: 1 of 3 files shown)


ivantha:
### Automated review summary

**Triage:** enhancement  |  **Recommendation:** ready-for-maintainer

**Findings**
- **[minor/test]** `tensormap-frontend/src/components/ErrorBoundary.test.jsx:97` The 'recovers after retry resets state' test is tautological and gives false confidence. It does `new ErrorBoundary({ children: null })` to create an instance React never mounts, so the instance's updater is React's no-op ReactNoopUpdateQueue. As a result `ErrorBoundaryInstance.setState({ hasError: true, error: new Error('test') })` does NOT mutate `this.state` (and logs a warning); `this.state` stays at the constructor value `{ hasError: false, error: null }`. handleRetry() then also no-ops. The final assertions `expect(hasError).toBe(false)` / `expect(error).toBeNull()` pass only because they happen to match the untouched initial state, so the test would still pass even if handleRetry were deleted. It never exercises the true->false retry transition it claims to verify. Drive retry through a mounted component instead (e.g. render a child that throws, assert the fallback, then click 'Try again' and assert recovery).
- **[minor/quality]** `tensormap-frontend/src/App.jsx:22` Coverage regression for the layout shell. The PR removes the single outer <ErrorBoundary> that previously wrapped the entire <Routes> tree and replaces it with per-route boundaries applied only to the five leaf pages via withErrorBoundary. The parent route element={<WorkspaceLayout />} is NOT wrapped, WorkspaceLayout has no boundary of its own (it just renders <Outlet/> plus sidebar/breadcrumb), and main.jsx has no root error boundary. So a render error thrown by WorkspaceLayout, AppTopBar, or the routing/Navigate machinery is now uncaught and unmounts the whole React tree (blank screen), whereas before it was caught and showed the fallback. Recommend keeping an outer boundary in addition to the per-route ones (or wrapping WorkspaceLayout) so the shared shell still has a safety net.
- **[nit/quality]** `tensormap-frontend/src/App.jsx:13` withErrorBoundary returns a fresh JSX element built inline and is invoked directly in each route's `element` prop, so a new <ErrorBoundary>/<Component> element tree is constructed on every render of App. App does not re-render in practice (it sits under RecoilRoot at the root), so there is no real-world remount cost today, but a helper that returns an element on each call is a fragile pattern. Consider memoizing the wrapped elements or having withErrorBoundary return a stable component (HOC) rather than an element.

PR #346 is a clean, self-contained frontend resilience/UX enhancement (per-route error boundaries, an ErrorFallback with retry + 'Back to Projects' navigation, and a PageLoader spinner) that does not overlap the reserved GSoC #120 scope at all, so it is safe to consider on its own merits. Two things to address before merge: (1) the 'recovers after retry resets state' test (ErrorBoundary.test.jsx:97) is tautological — it manually constructs an unmounted ErrorBoundary whose setState is a no-op, so it would pass even if handleRetry were removed; rewrite it to drive retry through a mounted component. (2) Replacing the single top-level boundary with leaf-only per-route boundaries leaves WorkspaceLayout, AppTopBar, and the router machinery with no boundary above them (no root boundary exists in main.jsx), a regression where a shell-level crash now blanks the whole app; keep an outer boundary in addition to the per-route ones. Note: I could not execute the suite locally — every test file fails to start due to an unrelated jsdom/html-encoding-sniffer ESM-require incompatibility with the installed Node 22.11/vitest 4 toolchain, not anything in this PR.

shivv23:

**Minor: the "recovers after retry resets state" test is tautological**
- Rewrote the test to mount a real `ErrorBoundary` with a `FlakyComponent` that conditionally throws. After the fallback appears, the component is fixed and `userEvent.click` drives the retry button, proving the true-to-false state transition actually works through React's lifecycle.

**Minor: WorkspaceLayout and AppTopBar lose the outer boundary**
- Kept the per-route boundaries and added an outer `<ErrorBoundary>` wrapping the entire `<Routes>` block inside `<Suspense>`. A crash in `WorkspaceLayout`, `AppTopBar`, or the router machinery is now caught.

**Nit: withErrorBoundary returns inline JSX, not memoized**
- Converted `withErrorBoundary` from an inline element factory to a proper HOC that returns a stable wrapper component with a `displayName` for easier debugging.

ivantha:
Please fix the issue in test cases.

shivv23:
Fixed the 'frontend-tests' failure: ErrorBoundary.test.jsx was importing @testing-library/user-event (not in dependencies). Replaced with fireEvent from the already-installed @testing-library/react — all 54 tests pass locally and CI is green.

PR #347, opened 2026-05-22
feat: add request ID middleware for request tracing

shivv23:
## Summary

Add a `RequestIDMiddleware` that attaches a unique `X-Request-ID` header to every HTTP response. This enables request tracing across logs and client-side debugging.

## Changes

- **`app/middleware.py`**: Added `RequestIDMiddleware` — generates a UUID if none is provided, preserves client-sent IDs
- **`app/main.py`**: Registered `RequestIDMiddleware` before `RequestLoggingMiddleware`
- **4 new tests**: validates header presence, UUID format, client value preservation, and uniqueness per request

## Validation

- 4 new unit tests added

### DIFF
--- a/tensormap-backend/app/main.py
+++ b/tensormap-backend/app/main.py
@@ -51,8 +51,8 @@ app.add_middleware(
     allow_methods=["*"],
     allow_headers=["*"],
 )
-app.add_middleware(RequestIDMiddleware)
 app.add_middleware(RequestLoggingMiddleware)
+app.add_middleware(RequestIDMiddleware)
 
 app.add_exception_handler(AppException, app_exception_handler)
 app.add_exception_handler(IntegrityError, integrity_error_handler)
--- a/tensormap-backend/app/main.py
+++ b/tensormap-backend/app/main.py
@@ -20,7 +20,7 @@ from app.exceptions import (
     integrity_error_handler,
     validation_exception_handler,
 )
-from app.middleware import RequestLoggingMiddleware
+from app.middleware import RequestIDMiddleware, RequestLoggingMiddleware
 from app.routers import data_process, data_upload, deep_learning, health, project
 from app.shared.logging_config import get_logger
 from app.socketio_instance import sio
@@ -51,6 +51,7 @@ app.add_middleware(
     allow_methods=["*"],
     allow_headers=["*"],
 )
+app.add_middleware(RequestIDMiddleware)
 app.add_middleware(RequestLoggingMiddleware)
 
 app.add_exception_handler(AppException, app_exception_handler)
--- a/tensormap-backend/app/middleware.py
+++ b/tensormap-backend/app/middleware.py
@@ -1,7 +1,8 @@
-"""HTTP request/response logging middleware."""
+"""HTTP request/response logging and tracing middleware."""
 
 import re
 import time
+import uuid
 
 from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.requests import Request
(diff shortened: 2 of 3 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/quality]** `tensormap-backend/app/middleware.py:19` The PR's stated purpose is not delivered. The PR description and the new docstring say the middleware 'enables request tracing across logs', but the request ID is never fed into the logging system. app/shared/logging_config.py uses a fixed formatter ('%(asctime)s - %(name)s - %(levelname)s - %(message)s') with no request-id field, and RequestLoggingMiddleware logs method/path/status/duration without referencing the ID. There is no contextvar, LogRecord filter, or formatter change wiring the ID into log lines, so it appears only as a response header. Add a contextvar set in dispatch plus a logging filter/formatter field, or scope the PR/description to 'response header only'.
- **[minor/bug]** `tensormap-backend/app/main.py:54` Middleware ordering disconnects the ID from request logs. In Starlette/FastAPI add_middleware prepends, so the last-added middleware is outermost. RequestIDMiddleware is added immediately before RequestLoggingMiddleware, making runtime order (outer to inner) CORS -> RequestLoggingMiddleware -> RequestIDMiddleware. RequestLoggingMiddleware therefore wraps the ID middleware and cannot see the generated/forwarded ID even if logging were wired to read it. To correlate logs with the header, RequestIDMiddleware must be the outer one (added after RequestLoggingMiddleware).
- **[minor/security]** `tensormap-backend/app/middleware.py:23` Client-supplied X-Request-ID is reflected verbatim with no validation, length cap, or charset check: request_id = request.headers.get('X-Request-ID', str(uuid.uuid4())) is written straight back into the response header, and test_request_id_preserves_client_value locks this echo in. CR/LF response-splitting is blocked by the ASGI server, but echoing unbounded untrusted input as a trace identifier becomes a log-injection/forging vector the moment the ID is written into logs (the stated goal). Validate the incoming value (e.g. require UUID format or cap length/charset) before reusing it.
- **[nit/test]** `tensormap-backend/tests/test_logging_middleware.py:96` test_request_id_is_valid_uuid is a weak format check. all(len(p) in (8, 4, 4, 4, 12) for p in parts) only asserts each of the 5 segments has a length drawn from the set {8,4,4,4,12}, not that segment lengths match the positional 8-4-4-4-12 UUID pattern, so a malformed id with the wrong per-position lengths can pass. Assert uuid.UUID(request_id) instead.

PR #347 adds a small, self-contained RequestIDMiddleware that stamps an X-Request-ID header on every response and includes four unit tests; it has no overlap with the reserved GSoC #120 scope, so it is safe to consider on its own merits. The main gap is that it does not actually achieve its stated goal of 'request tracing across logs' - the ID is only a response header and is never injected into the logger (no contextvar/formatter change), and the middleware ordering even places the logging middleware outside the ID middleware. Before merge, either wire the ID into the logging config (and reorder so RequestIDMiddleware is outermost) or narrow the description to 'response header only', and validate/bound the client-supplied X-Request-ID since it is reflected verbatim. The UUID-format test should also use uuid.UUID() rather than counting segment lengths.

shivv23:

**Major: request ID not wired into logging system**
- Added a `request_id_var` (`ContextVar[str]`) set in `RequestIDMiddleware.dispatch` and a `RequestIDFilter` registered on the root logger at import time. Every log record now carries a `request_id` field.
- `RequestLoggingMiddleware` includes `request_id` in its log line: `[request_id=<uuid>]`.

**Minor: middleware ordering reversed**
- Swapped the `add_middleware` order in `main.py`: `RequestLoggingMiddleware` is added first (inner), `RequestIDMiddleware` second (outer), so the ID is available before logging runs.

**Security: client-supplied X-Request-ID reflected verbatim**
- Incoming `X-Request-ID` is only accepted when it matches UUID format via `_UUID_RE.fullmatch()`; otherwise a fresh UUID is generated. CR/LF injection is already blocked by the ASGI server, and log injection is prevented by the UUID-only gate.

**Nit: weak UUID format check in tests**
- Replaced the segment-length-counting check with `uuid.UUID()` parsing in `test_request_id_is_valid_uuid`.
- Added `test_invalid_client_id_is_overridden` to verify non-UUID values are replaced.
- Added `test_request_id_in_logs` to verify IDs appear in log output.

PR #348, opened 2026-05-22
feat: add rate limiter utility for API endpoints

shivv23:
## Summary

Add a token-bucket rate limiter utility and optional middleware for API endpoint protection.

## Changes

- **`app/rate_limiter.py`**: `RateLimiter` class (IP-based, configurable RPM/window) + `RateLimitMiddleware` for easy integration
- **9 unit tests**: covering under-limit, over-limit, window expiry, X-Forwarded-For, multi-client isolation, and middleware integration

## Validation

- 9 new unit tests added

### DIFF
--- a/tensormap-backend/app/rate_limiter.py
+++ b/tensormap-backend/app/rate_limiter.py
@@ -1,0 +1,58 @@
+"""Rate limiting utilities for API endpoints."""
+
+import time
+from collections import defaultdict
+
+from fastapi import HTTPException, Request
+from starlette.middleware.base import BaseHTTPMiddleware
+
+from app.shared.logging_config import get_logger
+
+logger = get_logger(__name__)
+
+
+class RateLimiter:
+    """Token bucket rate limiter keyed by client IP."""
+
+    def __init__(self, requests_per_minute: int = 60, window_seconds: int = 60):
+        self.requests_per_minute = requests_per_minute
+        self.window_seconds = window_seconds
+        self._clients: dict[str, list[float]] = defaultdict(list)
+
+    def _get_client_id(self, request: Request) -> str:
+        forwarded = request.headers.get("X-Forwarded-For")
+        if forwarded:
+            return forwarded.split(",")[0].strip()
+        client_host = request.client.host if request.client else "unknown"
+        return client_host
+
+    def check(self, request: Request) -> None:
+        """Raise HTTPException 429 if the client has exceeded the rate limit."""
+        client_id = self._get_client_id(request)
+        now = time.time()
+        window_start = now - self.window_seconds
+
+        timestamps = self._clients[client_id]
+        timestamps[:] = [t for t in timestamps if t > window_start]
+
+        if len(timestamps) >= self.requests_per_minute:
+            logger.warning("Rate limit exceeded for client %s", client_id)
+            raise HTTPException(status_code=429, detail="Rate limit exceeded. Please slow down.")
+
+        timestamps.append(now)
+
+
+class RateLimitMiddleware(BaseHTTPMiddleware):
+    """Middleware that applies rate limiting to all incoming requests."""
+
+    def __init__(self, app, requests_per_minute: int = 60, window_seconds: int = 60):
+        super().__init__(app)
+        self.limiter = RateLimiter(
+            requests_per_minute=requests_per_minute,
+            window_seconds=window_seconds,
+        )
+
+    async def dispatch(self, request: Request, call_next):
+        self.limiter.check(request)
+        response = await call_next(request)
+        return response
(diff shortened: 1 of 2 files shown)


ivantha:
### Automated review summary

**Findings**
- **[blocking/bug]** `tensormap-backend/app/rate_limiter.py:50` RateLimitMiddleware.dispatch calls self.limiter.check(request), which raises fastapi.HTTPException from inside a BaseHTTPMiddleware. Starlette's ExceptionMiddleware that converts HTTPException into a response sits BELOW the user middleware stack, so an HTTPException raised in dispatch is never caught and converted. I reconstructed both files and ran them: wired into a real FastAPI app (which app/main.py is), the third request returns HTTP 500, not 429. The middleware therefore does not rate-limit as advertised. Fix: return a starlette JSONResponse(status_code=429, ...) directly from dispatch instead of raising, or move the check into a per-route dependency where raising HTTPException is valid.
- **[blocking/test]** `tensormap-backend/tests/test_rate_limiter.py:119` The PR's own bundled test test_middleware_blocks_excessive_requests FAILS as written. I ran the exact reconstructed test file: 6 of 7 tests pass, but this one fails because the raised HTTPException propagates out of BaseHTTPMiddleware on the third request instead of producing a 200-then-429 response, so 'assert response.status_code == 429' never executes (the exception surfaces first). This proves the suite was not run green before submission and the same defect breaks the feature in production.
- **[minor/security]** `tensormap-backend/app/rate_limiter.py:29` _get_client_id unconditionally trusts the X-Forwarded-For header and uses its first token as the rate-limit key. Any client can spoof an arbitrary X-Forwarded-For value per request to get a fresh bucket and bypass the limit entirely. Only honor this header when the request is known to come from a trusted proxy (e.g. a configurable trusted-proxy allowlist), otherwise fall back to request.client.host.
- **[minor/quality]** `tensormap-backend/app/rate_limiter.py:19` Unbounded memory growth and no cross-worker sharing: self._clients is an in-process defaultdict(list) keyed by client IP. Keys for idle clients are never evicted (only timestamps within a key are pruned on access), so a stream of distinct IPs grows the dict without bound. State also lives per Uvicorn worker, so limits are not enforced consistently under multiple workers. For real protection this needs a shared store (e.g. Redis) and idle-key eviction; at minimum document the single-process limitation.
- **[nit/quality]** `tensormap-backend/app/rate_limiter.py:12` Docstrings on both RateLimiter and the module call this a 'Token bucket' rate limiter, but the implementation is a fixed/sliding-window request log (it prunes timestamps older than window_seconds and counts the remainder). There is no token-refill or burst-capacity logic. Rename to 'sliding-window' to avoid misleading future maintainers.

PR #348 adds a standalone IP-based rate limiter (app/rate_limiter.py) plus 7 unit tests; it is not currently wired into app/main.py and has zero overlap with the reserved GSoC scope in issue #120, so it is safe from a scope-fencing standpoint. However, it should not be merged as-is: it has a blocking correctness bug. RateLimitMiddleware raises fastapi.HTTPException from inside a BaseHTTPMiddleware.dispatch, which Starlette does not convert to a response, so when used in the real FastAPI app the over-limit case returns 500 instead of 429 (I verified this by reconstructing and running the code). As a direct consequence the PR's own test_middleware_blocks_excessive_requests fails, showing the suite was not run green before submission. Ask the contributor to return a JSONResponse(429) from dispatch (or use a route dependency), then also address the minor issues: X-Forwarded-For is trusted unconditionally (rate-limit bypass), per-worker in-memory state grows unbounded with no idle-key eviction, and the 'token bucket' docstrings actually describe a sliding-window limiter.

shivv23:

**Blocking: HTTPException inside BaseHTTPMiddleware.dispatch returns 500 instead of 429**
- Changed `RateLimiter.check()` to raise a custom `RateLimitExceeded` exception instead of `fastapi.HTTPException`.
- `RateLimitMiddleware.dispatch` catches `RateLimitExceeded` and returns `starlette.responses.JSONResponse(status_code=429)` directly, since Starlette's `ExceptionMiddleware` does not catch exceptions raised inside `BaseHTTPMiddleware`.

**Blocking: test_middleware_blocks_excessive_requests fails**
- The test now correctly receives a 429 response from the middleware, passing as expected.

**Security: X-Forwarded-For spoofing**
- Added a `trusted_proxies` parameter. The `X-Forwarded-For` header is only honoured when the immediate peer is in the trusted proxies set; otherwise falls back to `request.client.host`.

**Quality: unbounded memory growth**
- Added periodic stale-key pruning (`_prune_stale_keys`), triggered every 100 `check()` calls.

**Nit: misleading "token bucket" docstrings**
- Corrected to "sliding-window rate limiter" throughout.

shivv23:
Further fix: corrected local import ordering within the third-party import block (RateLimiter, RateLimitExceeded, RateLimitMiddleware sorted case-insensitively). Backend Lint now passes.

PR #349, opened 2026-05-22
fix: validate target column exists in dataset and reject non-CSV files

shivv23:
## Summary

Improve robustness of the target field assignment endpoint by validating that the target column actually exists in the dataset, and rejecting non-CSV file types.

## Changes

- **`add_target_service`**: Now returns 400 if the file is not CSV, and validates the target column exists in `file.columns` before creating the record
- **3 new tests**: covers non-CSV rejection, missing column rejection, and valid column acceptance

## Validation

- 3 new unit tests added

### DIFF
--- a/tensormap-backend/app/services/data_process.py
+++ b/tensormap-backend/app/services/data_process.py
@@ -42,6 +42,14 @@ def add_target_service(db: Session, file_id: uuid_pkg.UUID, target: str) -> tupl
         file = db.exec(select(DataFile).where(DataFile.id == file_id)).first()
         if not file:
             return _resp(400, False, "File doesn't exist in DB")
+        if file.file_type != "csv":
+            return _resp(400, False, "Only CSV files support target field assignment")
+        if file.columns is not None and target not in file.columns:
+            return _resp(
+                400,
+                False,
+                f"Target column '{target}' not found in dataset columns",
+            )
 
         data_process = DataProcess(file_id=file_id, target=target)
         db.add(data_process)
--- a/tensormap-backend/tests/test_data_process_service.py
+++ b/tensormap-backend/tests/test_data_process_service.py
@@ -110,6 +110,45 @@ class TestAddTargetService:
         assert body["success"] is False
         assert "doesn't exist" in body["message"]
 
+    def test_rejects_non_csv_file(self, mock_db, file_id):
+        zip_file = MagicMock(spec=DataFile)
+        zip_file.id = file_id
+        zip_file.file_type = "zip"
+        mock_db.exec.return_value.first.return_value = zip_file
+
+        body, status = add_target_service(mock_db, file_id, "species")
+
+        assert status == 400
+        assert body["success"] is False
+        assert "CSV" in body["message"]
+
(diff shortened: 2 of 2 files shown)


ivantha:
### Automated review summary

**Findings**
- **[blocking/test]** `tensormap-backend/tests/test_data_process_service.py:94` This change regresses the pre-existing test_success test (verified by running `uv run pytest tests/test_data_process_service.py` on the PR branch: 1 failed, 28 passed; failure is test_success asserting `400 == 201`). The new guard `if file.columns is not None and target not in file.columns` breaks it: test_success uses the `sample_file` fixture, a `MagicMock(spec=DataFile)` that never sets `.columns`. A spec'd MagicMock returns a child MagicMock (NOT None) for an unset attribute, so `file.columns is not None` is True, and `'species' not in <MagicMock>` evaluates to True via MagicMock.__contains__, so the service returns 400 instead of 201. The PR author added/updated only the new tests and did not update the existing fixture/test, so CI on this repo's pytest suite will fail. Fix: set `.columns = None` (or a list containing the target) on the `sample_file` fixture at tests/test_data_process_service.py:40-47, or have the fixture include the target column.
- **[minor/quality]** `tensormap-backend/app/services/data_process.py:47` The column-existence check is silently skipped whenever `file.columns is None`. data_upload.py only caches columns for CSVs whose header parses successfully; on a ParserError/OSError/UnicodeDecodeError/MemoryError it logs a warning and leaves columns=None (data_upload.py:60-66). For such files this guard does nothing, so the 'validate target column exists' guarantee the PR title advertises is best-effort, not enforced. This fallback path is neither tested nor documented in the PR. Consider either documenting the None-means-skip behavior or treating an unparsed CSV as an error.

PR #349 adds two reasonable, well-scoped server-side validations to add_target_service (reject non-CSV files; reject a target not present in the cached DataFile.columns), and the change is grounded in real behavior since columns are populated for CSVs at upload (data_upload.py:60-74). It does NOT touch the reserved GSoC scope in issue #120, so there is no scope-fencing problem. However it must not merge as-is: it regresses the existing test_success unit test (confirmed by running pytest on the branch: 1 failed, 28 passed), because the pre-existing sample_file fixture is a spec'd MagicMock with no `.columns` set, which the new `file.columns is not None` guard treats as a non-None mock and rejects with 400. Ask the author to fix the fixture/test (set columns=None or include the target) and re-run the suite; the new code lints and formats cleanly otherwise.

shivv23:

**Blocking: test_success regresses to 400 instead of 201**
- Set `f.columns = None` on the `sample_file` fixture so the new `file.columns is not None` guard evaluates to `False` and the test proceeds to return 201 as expected.
- The new tests (`test_rejects_non_csv_file`, `test_rejects_column_not_in_dataset`, `test_accepts_column_in_dataset`) already cover the guard's happy and sad paths.

**Minor: columns=None path silently skipped**
- The fallback when `file.columns is None` is intentional: files whose columns could not be parsed at upload time skip the column-existence check rather than failing. This behaviour is consistent with how `data_upload.py` handles unparseable CSVs.

PR #266, opened 2026-03-26
Feat: Implement LSTM, GlobalAveragePooling2D, and DepthwiseConv2D Layers

Ojas2095:
Implemented three new Keras layers to TensorMap. The changes include adding the nodes properties, sidebar buttons, and frontend components as well as their Python generation code in the backend.

PR #350, opened 2026-05-29
Feat: Implement DepthwiseConv2D Layer

Ojas2095:
## Summary

This PR implements the DepthwiseConv2D layer as an isolated, rebase-fresh PR, superseding the DepthwiseConv2D part of #266.

### Changes
- Created `DepthwiseConvNode` component and added it to the sidebar/canvas.
- Configurable parameters: Kernel size (X,Y), Stride (X,Y), Padding, and Activation.
- Backend code generation in `model_generation.py` handles the new layer configuration.

### DIFF
--- a/tensormap-backend/app/services/model_generation.py
+++ b/tensormap-backend/app/services/model_generation.py
@@ -120,5 +120,15 @@ def _build_layer(node: dict, input_tensor):
             raise ValueError(f"Dropout rate must be in [0, 1), got {rate!r}.")
         return tf.keras.layers.Dropout(rate=rate, name=name)(input_tensor)
 
+    elif node_type == "customdepthwiseconv":
+        activation = params.get("activation", "none")
+        return tf.keras.layers.DepthwiseConv2D(
+            kernel_size=(int(params.get("kernelX", 3)), int(params.get("kernelY", 3))),
+            strides=(int(params.get("strideX", 1)), int(params.get("strideY", 1))),
+            padding=params.get("padding", "same"),
+            activation="linear" if activation == "none" else activation,
+            name=name,
+        )(input_tensor)
+
     else:
         raise ValueError(f"Unknown node type: {node_type}")
--- a/tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx
+++ b/tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx
@@ -41,6 +41,7 @@ import { trainingHistory as trainingHistoryAtom } from "../../shared/atoms";
 import ContextMenu from "./ContextMenu";
 import useUndoRedo from "../../hooks/useUndoRedo";
 import GlobalAvgPoolNode from "./CustomNodes/GlobalAvgPoolNode/GlobalAvgPoolNode";
+import DepthwiseConvNode from "./CustomNodes/DepthwiseConvNode/DepthwiseConvNode";
 
 const isMac =
   typeof navigator !== "undefined"
@@ -57,6 +58,7 @@ const nodeTypes = {
   customdropout: DropoutNode,
   custommaxpool: MaxPoolingNode,
   customglobalavgpool: GlobalAvgPoolNode,
+  customdepthwiseconv: DepthwiseConvNode,
 };
 
 // Keys MUST match nodeTypes above. Add a description when adding a node type.
@@ -68,6 +70,7 @@ const nodeDescriptions = {
   customdropout: "Randomly drops a fraction of inputs during training to reduce overfitting.",
   custommaxpool: "Downsamples by taking the maximum value in each pooling window.",
   customglobalavgpool: "Averages each feature map down to a single value.",
+  customdepthwiseconv: "Applies a depthwise convolution filter.",
 };
 
 const getTooltipStyle = (x, y) => {
@@ -604,6 +607,14 @@ function Canvas() {
         customdropout: { rate: "" },
         custommaxpool: { pool_size: "", stride: "", padding: "valid" },
         customglobalavgpool: {},
+        customdepthwiseconv: {
+          kernelX: "",
+          kernelY: "",
+          strideX: "",
+          strideY: "",
+          padding: "same",
+          activation: "none",
+        },
       };
 
       const newNode = {
--- a/tensormap-frontend/src/components/DragAndDropCanvas/CustomNodes/DepthwiseConvNode/DepthwiseConvNode.jsx
+++ b/tensormap-frontend/src/components/DragAndDropCanvas/CustomNodes/DepthwiseConvNode/DepthwiseConvNode.jsx
@@ -1,0 +1,38 @@
+import PropTypes from "prop-types";
+import { Handle, Position } from "reactflow";
+
+function DepthwiseConvNode({ data, id }) {
+  const { kernelX, kernelY, activation } = data.params;
+  return (
+    <div className="w-44 rounded-lg border bg-white shadow-sm">
+      <Handle type="target" position={Position.Left} isConnectable id={`${id}_in`} />
+      <div className="rounded-t-lg bg-node-depthwiseconv px-3 py-1.5 text-xs font-bold text-white">
+        DepthwiseConv2D
+      </div>
+      <div className="px-3 py-2 text-xs text-muted-foreground space-y-0.5">
+        {kernelX !== undefined && kernelX !== "" ? (
+          <div>
+            Kernel: {kernelX}×{kernelY}
+          </div>
+        ) : (
+          <div>Not configured</div>
+        )}
+        {activation && <div>Act: {activation}</div>}
+      </div>
+      <Handle type="source" position={Position.Right} isConnectable id={`${id}_out`} />
+    </div>
+  );
+}
+
+DepthwiseConvNode.propTypes = {
+  data: PropTypes.shape({
+    params: PropTypes.shape({
+      kernelX: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+      kernelY: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+      activation: PropTypes.string,
+    }).isRequired,
+  }).isRequired,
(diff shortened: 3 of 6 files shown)


ivantha:
### Automated review summary

**Triage:** enhancement  |  **Recommendation:** needs-coordination

> Please coordinate with the maintainers before continuing. This is unlikely to be merged as-is.

**Findings**
- **[blocking/bug]** `tensormap-frontend/src/components/DragAndDropCanvas/NodePropertiesPanel.jsx:313` The new MaxPooling2D properties form calls updateParam("pool_size", Number(...)) (and likewise for stride/padding), but this file redefines updateParam as a curried factory `const updateParam = (params, onNodeUpdate, id) => (name, value) => {...}`. Calling it with (name, value) binds params="pool_size" and onNodeUpdate=<number>, returns an inner function that is never invoked, so the field never updates and would call onNodeUpdate(id,...) on a number if it ever ran. The MaxPooling2D panel is non-functional. Use the already-defined doUpdate helper, e.g. onChange={(e) => doUpdate("pool_size", Number(e.target.value))}.
- **[blocking/bug]** `tensormap-frontend/src/components/DragAndDropCanvas/CustomNodes/DepthwiseConvNode/DepthwiseConvNode.jsx:7` DepthwiseConvNode uses className "bg-node-depthwiseconv" and the new Sidebar entry uses "border-l-node-depthwiseconv", but tailwind.config.js defines no node-depthwiseconv color (only node-input/dense/flatten/conv/dropout). The class resolves to nothing, so the node header renders with no background color. Add a node-depthwiseconv entry to the theme.extend.colors block.
- **[major/bug]** `tensormap-frontend/src/components/DragAndDropCanvas/Sidebar.jsx:59` Sidebar.jsx contains two identical MaxPooling2D draggable entries, both wired to onDragStart(e, "custommaxpool"). One is a copy-paste duplicate and should be removed.
- **[major/quality]** `tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx:1` Four files (model_generation.py, Canvas.jsx, NodePropertiesPanel.jsx, Sidebar.jsx) are re-encoded from LF to CRLF line endings (0 carriage-return lines at the branch point vs 134/795/448/81 in this branch). This rewrites every line, inflating the diff to ~1496 additions / 1111 deletions for what is otherwise a ~10-line feature, hiding the real change and poisoning git blame. Re-save these files with LF endings so only the substantive lines appear in the diff.
- **[major/quality]** `tensormap-frontend/src/components/DragAndDropCanvas/Canvas.jsx:34` This branch was cut from an older base and re-introduces work already present on main: the models->trainingHistory atom swap, setTrainingHistory(modelObjects), the right-click delete-node flow, and the node-description hover tooltip. It also re-adds MaxPooling2D and GlobalAveragePooling2D, and Canvas.jsx imports MaxPoolingNode and GlobalAvgPoolNode but those component files are absent from this branch's tree (only DepthwiseConvNode.jsx is included), so the frontend will fail to resolve those imports as-is. Rebase onto current main so the PR contains only the DepthwiseConv2D-specific changes and the missing imports resolve.
- **[minor/quality]** `tensormap-frontend/src/components/DragAndDropCanvas/NodePropertiesPanel.jsx:412` In the DepthwiseConv2D panel, every input's onChange recreates the helper inline as updateParam(params, onNodeUpdate, id)(...) on each render instead of using the doUpdate / handleChange helpers already created once at the top of the component (as the Conv2D/Dense panels do). Reuse doUpdate/handleChange for consistency and to avoid recreating the closure per keystroke.
- **[minor/quality]** `tensormap-backend/app/services/model_generation.py:126` Backend DepthwiseConv2D defaults activation to "relu" via params.get("activation", "relu"), but the frontend seeds the node's default activation param as "none". The two defaults disagree; on a node where activation was never set the generated layer would silently differ from what the UI implies. Align the backend default with the frontend ("none" -> linear) or vice versa.

Despite the title "Implement DepthwiseConv2D Layer," the PR also adds MaxPooling2D, GlobalAveragePooling2D, and a Dropout build branch, and re-encodes four files to CRLF, ballooning the diff to ~1496/1111 lines over what is really a small feature. There are two blocking issues: the new MaxPooling2D properties panel calls the curried updateParam factory with the wrong arguments so the fields are non-functional, and DepthwiseConvNode/Sidebar reference a node-depthwiseconv Tailwind color that is not defined, leaving the node header unstyled. The branch is also cut from a stale base, duplicating a MaxPooling2D sidebar entry and importing MaxPoolingNode/GlobalAvgPoolNode components that are not present in this branch's tree, which would break the frontend build. Recommend rebasing onto current main, restoring LF line endings, scoping the PR down to just DepthwiseConv2D, then fixing the MaxPool handler and adding the missing color token.

ivantha:
### Automated review summary

**Triage:** enhancement  |  **Recommendation:** needs-coordination
**GSoC scope:** overlaps reserved GSoC scope (issue #120, full)

> Please coordinate with the maintainers before continuing. This is unlikely to be merged as-is.

**Findings**
- **[major/bug]** `tensormap-backend/app/services/model_generation.py:125` int(params.get("kernelX", 3)) only substitutes the default when the key is absent, never when it holds an empty string. The properties panel binds the numeric inputs as value={params.kernelX ?? ""} and lets the user clear a field, persisting "" into params. When that reaches the backend, int("") raises ValueError and model generation crashes. The same applies to kernelY, strideX, and strideY. The .get(..., default) form here gives a false sense of safety because the default cannot fire on the empty-string case the UI can produce; either coerce empties to the default (e.g. int(params.get("kernelX") or 3)) or validate inputs before building the layer. _(unverified)_
- **[minor/test]** `tensormap-backend/app/services/model_generation.py:123` The PR adds a new backend layer-construction branch and a new frontend node component but ships no tests, even though the repo has pytest and Jest configured. There is no coverage asserting that a customdepthwiseconv node produces a valid DepthwiseConv2D in the generated model JSON, nor any rendering test for DepthwiseConvNode. Add at least a backend test for the new branch to lock in expected kernel/stride/padding/activation handling.
- **[nit/quality]** `tensormap-backend/app/services/model_generation.py:124` Keras DepthwiseConv2D exposes depth_multiplier (default 1), which is the parameter that most distinguishes it from a plain depthwise op, but neither the properties panel nor the backend exposes it, so users are silently locked to depth_multiplier=1. Consider surfacing it (defaulting to 1) for completeness, or note in the PR that it is intentionally omitted.

**Conventions**
- _Hygiene is clean: a single Conventional-Commits commit, a clear title, and a description that explains the change and references the superseded PR; consider linking the closing issue explicitly with a "Closes #" keyword._

This PR cleanly adds a DepthwiseConv2D layer end to end (sidebar drag source, node component, properties panel, backend code generation, and a theme color), with good developer hygiene: one Conventional-Commits commit and a meaningful description. The main blocker is a correctness bug: the backend uses int(params.get(\"kernelX\", 3)), whose default never fires for the empty strings the numeric inputs can produce, so clearing a kernel or stride field crashes model generation with int(\"\"); coerce empties to defaults or validate before building. Also add tests for the new backend branch and node, and consider exposing depth_multiplier since it is what makes DepthwiseConv2D distinct.

ivantha:
### Automated review summary

**Triage:** enhancement  |  **Recommendation:** needs-coordination
**GSoC scope:** overlaps reserved GSoC scope (issue #120, full)

> Please coordinate with the maintainers before continuing. This is unlikely to be merged as-is.

**Findings**
- **[major/bug]** `tensormap-backend/app/services/model_generation.py:126` int(params.get("kernelX", 3)) does not guard against an empty string. params.get("kernelX", 3) returns "" (not 3) when the key is present with an empty-string value, which is exactly the initial Canvas state. int("") raises ValueError at runtime when a user submits an unconfigured node.
- **[major/bug]** `tensormap-backend/app/services/model_generation.py:127` Same empty-string crash applies to strideX and strideY. int(params.get("strideX", 1)) and int(params.get("strideY", 1)) both crash with ValueError when the user has not filled in those fields.
- **[minor/quality]** `tensormap-frontend/src/components/DragAndDropCanvas/CustomNodes/DepthwiseConvNode/DepthwiseConvNode.jsx:13` The node summary card checks only kernelX to decide whether to show 'Not configured', but kernelX and kernelY are entered independently. If kernelX is filled and kernelY is left empty the card renders 'Kernel: 5x', showing a dangling multiplication sign.
- **[nit/quality]** `tensormap-frontend/src/components/DragAndDropCanvas/CustomNodes/DepthwiseConvNode/DepthwiseConvNode.jsx:13` strideX and strideY are not shown anywhere in the node summary card, so a user cannot tell at a glance what stride was configured without opening the properties panel. Other nodes that have strides (e.g. Conv2D) show them.
- **[nit/quality]** `tensormap-frontend/src/components/DragAndDropCanvas/CustomNodes/DepthwiseConvNode/DepthwiseConvNode.jsx:67` PropTypes for DepthwiseConvNode omits strideX and strideY even though both are part of the node's params shape and are rendered (or should be rendered) in the summary.

**Conventions**
- **PR title:** Title capitalises 'Feat' and 'Implement' mid-word, breaking the lowercase-after-colon Conventional Commits style (should be 'feat: implement DepthwiseConv2D layer').
- **Commits (2):** The second commit ('fix: register DepthwiseConvNode and formatting') is a fixup that should have been squashed into the first before opening the PR; publishing incremental 'fix formatting' commits inflates blame history.
- _Squash the fix commit into the feature commit and lowercase the PR title before merging._

The PR adds a DepthwiseConv2D layer with a new React node, properties panel, sidebar entry, Tailwind colour token, and backend code-generation branch -- all consistent with how existing layers are structured. However, the backend int() calls on kernelX, kernelY, strideX, and strideY will raise ValueError at runtime if a user submits the node without filling those fields, because params.get(key, default) returns the empty-string stored in state rather than the integer fallback. Before merging, add explicit empty-string guards (e.g. int(v) if v else default) for all four numeric params, and the node summary card should display stride values and fix the split kernelX/kernelY check to avoid a dangling 'x' character.

ivantha:
Closing as stale and also due to the overlap with GSoC scope.

PR #353, opened 2026-06-04
fix: use curried doUpdate for Dropout and MaxPooling param handlers

shivv23:
updateParam is a curried function (params, onNodeUpdate, id) => (name, value) => void. It is correctly unwrapped on line 90 as const doUpdate = updateParam(params, onNodeUpdate, id). However, the Dropout and MaxPooling sections call the outer function directly instead of the returned inner handler, so the parameter update is silently discarded.

Broken:
onChange={(e) => updateParam("rate", e.target.value)}
(calls 3-arg curry with 2 args, returned fn is never invoked)

Fixed:
onChange={handleChange("rate")}
(uses the same handleNumberChange helper as Dense/Conv/Input)

Also applies ?? "" null-coalescing on all value bindings to prevent uncontrolled-to-controlled React warnings.

### DIFF
--- a/tensormap-frontend/src/components/DragAndDropCanvas/NodePropertiesPanel.jsx
+++ b/tensormap-frontend/src/components/DragAndDropCanvas/NodePropertiesPanel.jsx
@@ -296,8 +296,8 @@ function NodePropertiesPanel({
               min="0"
               max="1"
               step="0.1"
-              value={params.rate}
-              onChange={(e) => updateParam("rate", e.target.value)}
+              value={params.rate ?? ""}
+              onChange={handleChange("rate")}
             />
           </div>
         </CardContent>
@@ -317,8 +317,8 @@ function NodePropertiesPanel({
               type="number"
               min="1"
               placeholder="Pool size"
-              value={params.pool_size}
-              onChange={(e) => updateParam("pool_size", Number(e.target.value))}
+              value={params.pool_size ?? ""}
+              onChange={handleChange("pool_size")}
             />
           </div>
           <div className="space-y-1">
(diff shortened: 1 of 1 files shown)


shivv23:
updateParam is a curried function that returns a handler (name, value) => void. The Dropout and MaxPooling sections were calling updateParam(...) directly with wrong arity instead of calling the pre-bound doUpdate(...) or handleChange(...). The returned inner function was silently discarded -- params never reached React state.

ivantha:
### Automated review summary

**Findings**
- **[nit/quality]** `tensormap-frontend/src/components/DragAndDropCanvas/NodePropertiesPanel.jsx:336` The MaxPooling padding Select uses doUpdate("padding", v) while every other Select in this file (Dense activation, Conv padding, Conv activation) uses doUpdate(..., v || null). The padding options are only 'valid'/'same' so v is never empty in practice, making this harmless, but it is stylistically inconsistent with the sibling Select handlers. Consider doUpdate("padding", v || null) for uniformity.
- **[minor/bug]** `tensormap-frontend/src/components/DragAndDropCanvas/CustomNodes/MaxPoolingNode/MaxPoolingNode.jsx:7` Routing these inputs through handleChange/updateParam now writes null (instead of "") when a field is cleared. The node display guards check only `!== "" && !== undefined`, which do not catch null, so after clearing a value the canvas node renders 'Pool: null | Stride: null' / 'Rate: null' instead of 'Not configured'. This is cosmetic (the save validator still blocks via !p.pool_size and isNaN(parseFloat(null))), but the display guards should be widened, e.g. `pool_size != null && pool_size !== ""`, to keep the 'Not configured' label correct.
- **[minor/bug]** `tensormap-frontend/src/components/DragAndDropCanvas/CustomNodes/DropoutNode/DropoutNode.jsx:14` Same null-vs-empty-string display gap for Dropout: the guard `rate !== "" && rate !== undefined` does not catch null, so a cleared rate now shows 'Rate: null' rather than 'Not configured'. Widen to `rate != null && rate !== ""` to match the new null representation produced by the corrected handler.

Solid, correct bug fix. The Dropout and MaxPooling2D inputs were calling the curried updateParam(name, value) directly, so the returned inner handler was never invoked and onNodeUpdate never fired, silently discarding every edit to those params; rewiring them to the existing handleChange/doUpdate (the same helpers Input/Dense/Conv use) is the right fix, and the added `?? \"\"` correctly resolves controlled-input warnings. One stylistic nit: MaxPooling's padding Select omits the `v || null` fallback that its sibling Selects use. One follow-up worth landing: clearing a field now stores null instead of \"\", and the MaxPoolingNode/DropoutNode \"Not configured\" guards only test for \"\"/undefined, so a cleared field renders 'null' on the canvas (save validation still blocks it, so this is cosmetic). Safe to merge; consider widening those two display guards to also treat null as unconfigured.

PR #355, opened 2026-06-04
fix: allow ZIP uploads for image datasets

shivv23:
Unblock ZIP image dataset uploads and make the image-classification pipeline functional end to end.

Changes:

1. Router (data_upload.py): ALLOWED_EXTENSIONS now accepts both csv and zip.
2. Frontend (DataUpload.jsx): File picker accept attribute now permits .csv and .zip.
3. Upload service (data_upload.py):
   - On ZIP save, extracts the archive to {upload_folder}/{disk_name_no_ext}/ with zip-slip protection (path traversal check on every member) and a 2 GB decompression-bomb limit.
   - Creates an ImageProperties record (image_size=224, batch_size=32, color_mode=rgb, label_mode=int) linked to the same file_id so training and code-generation paths do not raise ModelRunError("Image properties not found").
   - Delete service now cleans up the extraction directory via shutil.rmtree.
4. Path resolvers:
   - model_run.py:_helper_generate_file_location returns the extraction directory for ZIP files (disk_name without .zip extension) and uses disk_name for all other files.
   - code_generation.py:_file_location follows the same logic.
5. Tests: Updated test_null_checks.py to assert extraction directory paths and added a ZIP path test for the code_generation resolver.

### DIFF
--- a/tensormap-backend/app/routers/data_upload.py
+++ b/tensormap-backend/app/routers/data_upload.py
@@ -57,7 +57,7 @@ def upload_file(
             status_code=400,
             content={
                 "success": False,
-                "message": "Only CSV files are supported.",
+                "message": "Only CSV and ZIP files are supported.",
                 "data": None,
             },
         )
--- a/tensormap-backend/app/routers/data_upload.py
+++ b/tensormap-backend/app/routers/data_upload.py
@@ -15,7 +15,7 @@ from app.shared.logging_config import get_logger
 
 logger = get_logger(__name__)
 router = APIRouter(tags=["data-upload"])
-ALLOWED_EXTENSIONS = {"csv"}
+ALLOWED_EXTENSIONS = {"csv", "zip"}
 
 
 class _UploadFileWrapper:
--- a/tensormap-backend/app/services/code_generation.py
+++ b/tensormap-backend/app/services/code_generation.py
@@ -93,4 +93,6 @@ def _file_location(file_id, db: Session) -> str:
     file = db.exec(select(DataFile).where(DataFile.id == file_id)).first()
     if file is None:
         raise CodeGenerationError(f"Dataset file not found (id={file_id})")
+    if file.file_type == "zip":
+        return f"{settings.upload_folder}/{file.disk_name.rsplit('.', 1)[0]}"
     return f"{settings.upload_folder}/{file.disk_name}"
--- a/tensormap-backend/app/services/data_upload.py
+++ b/tensormap-backend/app/services/data_upload.py
@@ -1,6 +1,8 @@
 import contextlib
 import os
+import shutil
 import uuid as uuid_pkg
+import zipfile
 from typing import Any
 
 import pandas as pd
@@ -10,7 +12,7 @@ from sqlmodel import Session, select
 from werkzeug.utils import secure_filename
 
 from app.config import get_settings
-from app.models import DataFile
+from app.models import DataFile, ImageProperties
 from app.shared.logging_config import get_logger
 
 logger = get_logger(__name__)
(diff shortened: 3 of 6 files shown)


shivv23:
The backend router's ALLOWED_EXTENSIONS = {"csv"} and the frontend's accept=".csv" both rejected ZIP files before they could reach the existing (and complete) ZIP handling code in add_file_service, code_generation.py, and model_run.py.

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `tensormap-backend/app/routers/data_upload.py:18` The PR description states the backend already has full ZIP support and this change merely unblocks the gate, but that premise does not hold end to end, so accepting a ZIP here exposes a non-functional path. The save path (add_file_service in app/services/data_upload.py) writes the archive to disk under disk_name = '{name}_{uuid8}.{ext}' (e.g. images_3f9a1c20.zip) while storing file_name = secure_filename(name) (e.g. 'images'). For ZIP files, the downstream resolvers return upload_folder + '/' + file.file_name (model_run.py line 134; code_generation.py line 97), i.e. '<folder>/images' with no extension and no uuid suffix. That path does not exist on disk and is not a directory, and tf.keras.utils.image_dataset_from_directory requires an existing directory of per-class subfolders. A ZIP upload will therefore save successfully but fail at code generation / training time. The two changed lines are themselves correct; the gap is in the unchanged resolve+extract logic the feature depends on.
- **[major/bug]** `tensormap-backend/app/services/data_upload.py:51` There is no ZIP extraction step anywhere in the backend (grep across app/ for ZipFile, zipfile, extractall, unpack_archive, .extract returns zero matches). The uploaded archive is stored as-is and never unpacked into a directory of class subfolders, which is exactly what image_dataset_from_directory(directory, ...) consumes. Enabling ZIP uploads without an extraction step means the image-classification dataset flow cannot complete.
- **[minor/test]** `tensormap-backend/app/routers/data_upload.py:55` The behavior change (router ALLOWED_EXTENSIONS now accepts 'zip', and frontend accept now permits '.zip') ships with no test. No test asserts that a .zip upload is accepted by POST /data/upload/file, nor that a non-allowed extension is still rejected. The existing upload-extension gate is otherwise untested for the new branch, so a future regression to {csv} would pass CI.
- **[minor/security]** `tensormap-backend/app/routers/data_upload.py:65` Permitting ZIP uploads introduces new attack surface that is currently unmitigated. The size cap enforced in upload_file only bounds the compressed archive, not the expanded contents (decompression-bomb risk), and once an extraction step is added it must guard against path traversal / zip-slip from crafted member names. No such protections exist yet. Not exploitable today because nothing extracts the archive, but this should be in place before extraction is implemented.

The two-line change (router allow-list adds 'zip'; frontend accept adds '.zip') is correct and minimal, and unblocking ZIP uploads is a reasonable goal. However, the PR description's claim that the backend already has full working ZIP support is not accurate end to end: nothing in the backend ever extracts the archive (no ZipFile/extractall anywhere), and for ZIP files the path resolvers in model_run.py (line 134) and code_generation.py (line 97) return '<upload_folder>/<file_name>' with no extension or uuid suffix, which does not match the saved disk_name and is not a directory, so image_dataset_from_directory will fail. As-is, a ZIP will upload successfully but training/code-gen for image classification cannot complete; please add an extraction step that unpacks into a per-class directory and align the ZIP path resolution with the actual on-disk name, ideally with a test covering the new accepted extension and zip-slip/size protections before merging.

shivv23:
Unblock ZIP image dataset uploads and make the image-classification pipeline functional end to end.

Changes:

1. Router (data_upload.py): ALLOWED_EXTENSIONS now accepts both csv and zip.
2. Frontend (DataUpload.jsx): File picker accept attribute now permits .csv and .zip.
3. Upload service (data_upload.py):
   - On ZIP save, extracts the archive to {upload_folder}/{disk_name_no_ext}/ with zip-slip protection (path traversal check on every member) and a 2 GB decompression-bomb limit.
   - Creates an ImageProperties record (image_size=224, batch_size=32, color_mode=rgb, label_mode=int) linked to the same file_id so training and code-generation paths do not raise ModelRunError("Image properties not found").
   - Delete service now cleans up the extraction directory via shutil.rmtree.
4. Path resolvers:
   - model_run.py:_helper_generate_file_location returns the extraction directory for ZIP files (disk_name without .zip extension) and uses disk_name for all other files.
   - code_generation.py:_file_location follows the same logic.
5. Tests: Updated test_null_checks.py to assert extraction directory paths and added a ZIP path test for the code_generation resolver.

ivantha:
### Automated review summary

**Triage:** enhancement  |  **Recommendation:** ready-for-maintainer

**Findings**
- **[major/security]** `tensormap-backend/app/services/data_upload.py:86` Zip-slip protection is name-based only and is evaluated before extraction, so it does not defend against symlink members. extractall() writes every member, including symlink entries whose names pass the os.path.realpath(...).startswith(extract_path + os.sep) check; a follow-up member that writes through such a symlink can still escape extract_path. Validate symlink members explicitly (reject them, or extract members one by one and re-check the resolved target after each write) instead of relying on a pre-pass name check plus a bulk extractall(). _(unverified)_
- **[minor/security]** `tensormap-backend/app/services/data_upload.py:79` The decompression-bomb guard sums zf.getinfo(member).file_size from the ZIP central directory rather than counting bytes actually written during extraction, so a member whose header understates its true size is not caught; additionally the 2 GB uncompressed ceiling is very large relative to the 200 MB compressed-upload cap enforced upstream, and there is no per-file or file-count limit. Prefer enforcing the limit on bytes streamed out during extraction.
- **[minor/quality]** `tensormap-backend/app/services/data_upload.py:103` ImageProperties is created with hardcoded image_size=224, batch_size=32, color_mode="rgb", label_mode="int" and there is no route or UI to change them, so every uploaded image dataset is silently forced to these preprocessing parameters. Consider accepting these as upload parameters (with these as defaults) so grayscale/non-224 datasets are usable.
- **[nit/quality]** `tensormap-backend/app/services/data_upload.py:86` The extraction directory is passed straight to tf.keras.utils.image_dataset_from_directory (via the path resolvers), which expects class subdirectories at its top level. A ZIP whose root is a single wrapping folder (a very common layout) would be interpreted as one class with no warning. Consider detecting/normalizing a single top-level directory, or validating that the extracted tree has the expected class-subdir structure and returning a clear error otherwise.

**Conventions**
- **Description:** Thorough and explains what/why, but does not reference or close any tracking issue.
- _Both commits are well-formed fix: messages; consider squashing the two same-scope fix: commits and adding a 'Closes #<n>' line to the description._

This PR enables ZIP image-dataset uploads end to end: it widens the allowed extensions, extracts archives with a path-traversal check and a 2 GB uncompressed cap, auto-creates an ImageProperties row so the image-classification path no longer errors, cleans up the extraction dir on delete, and updates both dataset-path resolvers and their tests. The main item to address before merge is the zip-slip protection: it is name-based and runs before a bulk extractall(), so symlink members can still escape the extraction directory; harden by rejecting symlinks or extracting and re-validating per member (data_upload.py:86). Secondary follow-ups: enforce the size cap on bytes actually written rather than trusting central-directory sizes (line 79), expose the currently hardcoded image preprocessing params (lines 103-110), and handle ZIPs that wrap their classes in a single top-level folder. Code structure, response handling, and test coverage are otherwise solid; just add a 'Closes #<issue>' reference to the description.

PR #358, opened 2026-06-08
fix: delete file from disk only after DB commit succeeds

shivv23:
In delete_one_file_by_id_service, files were removed from disk before the DB transaction committed. If the commit failed (e.g. FK constraint), the data was permanently lost even though the DB record survived.

### DIFF
--- a/tensormap-backend/app/services/data_upload.py
+++ b/tensormap-backend/app/services/data_upload.py
@@ -237,5 +237,6 @@ def delete_one_file_by_id_service(db: Session, file_id: uuid_pkg.UUID) -> tuple:
             shutil.rmtree(extract_path, ignore_errors=True)
     except OSError:
         logger.exception("Failed to remove disk file for id=%s", file_id)
+        return _resp(200, True, "File record deleted, but the physical file could not be removed from storage")
 
     return _resp(200, True, "File deleted successfully")
--- a/tensormap-backend/app/services/data_upload.py
+++ b/tensormap-backend/app/services/data_upload.py
@@ -12,7 +12,7 @@ from sqlmodel import Session, select
 from werkzeug.utils import secure_filename
 
 from app.config import get_settings
-from app.models import DataFile, ImageProperties
+from app.models import DataFile, ImageProperties, ModelBasic
 from app.shared.logging_config import get_logger
 
 logger = get_logger(__name__)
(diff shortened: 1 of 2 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `tensormap-backend/app/services/data_upload.py:230` After db.commit() succeeds, os.remove(file_path) is called. If it raises an OSError (e.g., permission error, not just FileNotFoundError), the outer except (SQLAlchemyError, OSError) block is reached, which logs the exception but returns nothing — the caller receives None instead of the expected tuple. The DB record is already deleted at that point, so the state is inconsistent and the caller likely crashes on None. The FileNotFoundError guard only catches one specific case; other OS errors fall through to the outer handler.
- **[major/bug]** `tensormap-backend/app/services/data_upload.py:216` model_basic.file_id is nulled out and committed unconditionally before deleting the DataFile. If the delete or commit fails, the ModelBasic record now permanently has a null file_id with no rollback, breaking referential integrity from the application's perspective. A db.rollback() should be issued (or the session used as a context manager) inside the except block so that the nullification is also rolled back on failure. _(unverified)_
- **[minor/quality]** `tensormap-backend/app/services/data_upload.py:218` The fix handles only the first ModelBasic that references the file (db.exec(...).first()). If multiple ModelBasic rows share the same file_id, only one is nulled and the delete will still fail with an FK constraint violation. Use .all() and iterate, or rely on a cascade/SET NULL at the DB level.
- **[nit/test]** `tensormap-backend/app/services/data_upload.py:209` No test is added or updated. The bug being fixed (disk removal before commit) is straightforward to cover with a unit test that mocks db.commit to raise and asserts the file still exists on disk. Adding a regression test would prevent this class of ordering bug from recurring.

**Conventions**
- **Description:** The description explains the bug but does not link to any issue number (e.g., 'Closes #NNN'), making it impossible to trace back to the reported bug.
- _Add a 'Closes #NNN' line to the PR description._

The PR correctly moves disk removal to after the DB commit, fixing the core ordering bug. However, the fix introduces a new atomicity hazard: model_basic.file_id is nulled and committed before the file record is deleted, so a failure mid-way leaves the DB in a partially modified state with no rollback. Additionally, only the first ModelBasic referencing the file is handled, which will cause an FK violation if multiple rows share the same file_id. The OSError path after a successful commit also returns None to the caller instead of an error tuple. These three issues should be addressed before merging, and a regression test covering the failure scenario would be valuable.

ivantha:
### Automated review summary

**Findings**
- **[minor/quality]** `tensormap-backend/app/services/data_upload.py:239` A disk-removal OSError after a successful commit is logged but the function still returns HTTP 200. This is documented in the comment as intentional, but the caller receives a success response even when the physical file is stuck on disk. Consider whether a 207 or a warning field in the response body would better signal the partial-failure state to callers.
- **[nit/test]** `tensormap-backend/tests/test_data_upload_service.py:42` All test cases that exercise the two-query code path (first() for the file lookup, all() for ModelBasic) share the same mock_db.exec.return_value, so mock_db.exec.return_value.first and .all are configured on the same object. This works because MagicMock returns the same object for every exec() call, but it means the tests do not verify that the correct SELECT statement is passed to each exec() call. Splitting into a side_effect list would make the tests more faithful to the real call sequence.

**Conventions**
- **Description:** The description states the bug but does not link a closing issue (e.g. 'Closes #NNN'), making it harder to trace which issue this resolves.
- _Add a 'Closes #NNN' line to the PR description so the related issue is auto-closed on merge._

The fix correctly inverts the delete order so the DB commit completes before any disk removal, preventing the original data-loss scenario where a failed commit left the DB record intact while the file was already gone. The ModelBasic FK nulling is a necessary companion change and is tested. Two minor issues worth addressing: (1) when disk removal fails after a successful commit the function still returns 200 with no signal to the caller that the file is stranded on disk; and (2) the test mocks share a single exec() return value for both the file-lookup and ModelBasic queries, which works but reduces fidelity. The PR description should also include a 'Closes #NNN' reference.

PR #359, opened 2026-06-08
fix: clean up uploaded files and model JSONs on disk when deleting a project

shivv23:
delete_project_service cascade-deletes DB records but leaves uploaded CSV/ZIP files, extracted ZIP directories, and generated model JSON files on disk. This change collects the disk paths before the DB cascade, deletes the DB records, then removes the orphaned files.

### DIFF
--- a/tensormap-backend/app/services/project.py
+++ b/tensormap-backend/app/services/project.py
@@ -173,7 +173,11 @@ def delete_project_service(db: Session, project_id: uuid_pkg.UUID) -> tuple:
         logger.exception("Error deleting project")
         return _resp(500, False, "An error occurred while deleting the project")
 
-    # Remove uploaded files from disk.
+    # Disk cleanup runs after the DB commit so that a failed commit never
+    # orphans disk records. The trade-off: if the process is interrupted
+    # between commit and cleanup, orphaned files remain with no DB record
+    # to retry from. This is acceptable because the files are scoped to a
+    # project that has been intentionally deleted.
     n_files = 0
     for file in data_files:
         file_path = os.path.realpath(os.path.join(upload_folder, file.disk_name))
@@ -182,6 +186,7 @@ def delete_project_service(db: Session, project_id: uuid_pkg.UUID) -> tuple:
             continue
         with contextlib.suppress(FileNotFoundError):
             os.remove(file_path)
+            # Only files that were actually present and removed are counted.
             n_files += 1
 
         if file.file_type == "zip" and "." in file.disk_name:
--- a/tensormap-backend/app/services/project.py
+++ b/tensormap-backend/app/services/project.py
@@ -1,3 +1,6 @@
+import contextlib
+import os
+import shutil
 import uuid as uuid_pkg
 from datetime import UTC, datetime
 from typing import Any
@@ -6,10 +9,12 @@ from sqlalchemy import func
 from sqlalchemy.exc import SQLAlchemyError
 from sqlmodel import Session, select
 
+from app.config import get_settings
 from app.models.data import DataFile
 from app.models.ml import ModelBasic
 from app.models.project import Project
 from app.schemas.project import ProjectCreateRequest, ProjectUpdateRequest
+from app.shared.constants import MODEL_GENERATION_LOCATION, MODEL_GENERATION_TYPE
 from app.shared.logging_config import get_logger
 
 logger = get_logger(__name__)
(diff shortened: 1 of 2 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/security]** `tensormap-backend/app/services/project.py:170` No path-traversal guard on `file.disk_name`. The model-JSON path is validated with `json_path.startswith(model_json_dir + os.sep)`, but the data-file path is not. If `disk_name` ever contains `..` segments, `os.remove` would operate outside `upload_folder`. Disk names appear to be UUID-based today, but the missing guard makes the code inconsistent and fragile under future changes.
- **[minor/quality]** `tensormap-backend/app/services/project.py:175` The extracted-ZIP directory name is reconstructed with `file.disk_name.rsplit('.', 1)[0]`, which silently does nothing if the extraction service stored the directory under a different name convention. The assumption should be documented, or the stored extraction path should be persisted in the DB so it can be removed reliably.
- **[minor/test]** `tensormap-backend/app/services/project.py:152` No tests are added for the new disk-cleanup logic. The function now has meaningful filesystem side-effects (file removal, `shutil.rmtree`) that should be covered by at least one unit test with a mocked filesystem, including the path-traversal guard branch.

**Conventions**

The PR correctly fixes orphaned-file leaks on project deletion by collecting disk paths before the DB cascade, committing, then removing files. The logic is sound and the model-JSON path has a path-traversal guard. The data-file removal path lacks the equivalent guard -- `disk_name` is never validated against `upload_folder` before calling `os.remove`, which is inconsistent with the model-JSON defence and should be fixed. The ZIP extraction-directory reconstruction from the filename is fragile if the naming convention ever changes, and no tests cover the new filesystem side-effects.

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `tensormap-backend/app/services/project.py:152` Disk cleanup runs outside the try/except SQLAlchemyError block. Any unexpected exception from os.remove or shutil.rmtree (beyond FileNotFoundError, which is suppressed) will propagate as an unhandled exception to the caller. The outer except only catches SQLAlchemyError, so OS-level errors during cleanup produce an unhandled 500.
- **[major/bug]** `tensormap-backend/tests/test_project_service.py:91` The path-traversal guard for DataFile uses `file_path.startswith(upload_folder + os.sep)` but `os.path.realpath` on a directory always strips any trailing separator, so on a system where `settings.upload_folder` ends with a separator, `upload_folder + os.sep` would never produce a double-slash issue -- however `os.path.realpath` already normalizes this. The same guard on line 180 for the zip extraction dir is correct. The real gap is that a file sitting directly inside upload_folder (not in a subdirectory) will fail the guard: `realpath('/uploads/data.csv').startswith('/uploads' + '/')` is True only when the resolved path starts with `/uploads/`, which it does for a file directly inside. This is actually fine -- but the same test fixture patches MODEL_GENERATION_LOCATION with a trailing slash, meaning os.path.realpath is bypassed in the test, so the guard behavior tested differs from production. _(unverified)_
- **[minor/quality]** `tensormap-backend/app/services/project.py:183` If a DataFile.disk_name has no dot (e.g. a dotless filename stored without extension), `disk_name.rsplit('.', 1)[0]` returns the full name unchanged, making extract_dir_name equal to disk_name. The code would then attempt to rmtree a directory named identically to the file. This is unlikely in practice for zip files but the assumption is implicit and unguarded.
- **[nit/quality]** `tensormap-backend/app/services/project.py:168` Successful file removals produce no log output; only blocked traversal attempts are logged. Adding a debug-level log for each removed file and a summary count would improve observability during incident investigation.

**Conventions**

This PR correctly addresses the disk-leak bug by collecting DB records before the cascade delete and then removing the orphaned files, with path-traversal guards for both uploaded files and model JSONs. The main concern is that the disk-cleanup block sits outside the SQLAlchemyError try/except, so any unexpected OS error there (not FileNotFoundError) will propagate as an unhandled exception; wrapping the cleanup in its own broad except with a logger.exception call would make failures recoverable and visible. A secondary issue is that the test for MODEL_GENERATION_LOCATION patches in a trailing-slash path, bypassing the os.path.realpath normalization that production code performs, so the path-traversal guard for model JSONs is not exercised under realistic conditions. The rest of the implementation -- contextlib.suppress for missing files, realpath normalization, and zip extraction directory removal -- is solid and the test coverage is a meaningful addition.

ivantha:
### Automated review summary

**Findings**
- **[minor/bug]** `tensormap-backend/app/services/project.py:188` n_files is incremented inside the contextlib.suppress block, so if os.remove raises FileNotFoundError the count is silently skipped but the log message still runs later with potentially undercounted files. This is not wrong per se, but the counter name implies 'files removed', not 'files found'. A comment clarifying this intent would prevent future confusion.
- **[minor/test]** `tensormap-backend/tests/test_project_service.py:107` There is a path-traversal test for DataFile.disk_name but no equivalent test for a malicious ModelBasic.model_name value (e.g. '../../../etc/model'). The guard exists in the production code but is untested.
- **[minor/quality]** `tensormap-backend/app/services/project.py:160` The DB commit happens before any disk cleanup. If the process is interrupted after commit but before cleanup completes, the orphaned files have no remaining DB records and cannot be cleaned up automatically in a future run. A comment acknowledging this trade-off (and why it is acceptable) would help future readers.

**Conventions**
- **Commits (3):** The third commit is a merge commit ('Merge branch main into fix/project-cleanup-disk') that should be removed by rebasing onto main rather than merging, keeping the branch history linear.
- _Rebase the branch onto main and drop the merge commit before merging._

This PR correctly fixes the long-standing gap where deleting a project left uploaded CSV/ZIP files and generated model JSON files on disk. The path-traversal guards using os.path.realpath plus a prefix check are well-implemented, and the five new unit tests cover the main cases. Two minor gaps worth addressing before merge: add a path-traversal test for model_name to match the one already present for disk_name, and rebase the branch to drop the merge commit. The atomicity trade-off (DB commit before disk cleanup) is acceptable for a fix of this scope but deserves a brief inline comment.

PR #360, opened 2026-06-08
fix: back up original CSV before overwriting during preprocessing

shivv23:
preprocess_data calls df.to_csv(file_path, index=False) which overwrites the original uploaded CSV with no way to recover. This change creates a .bak copy via shutil.copy2 before the overwrite.

### DIFF
--- a/tensormap-backend/app/services/data_process.py
+++ b/tensormap-backend/app/services/data_process.py
@@ -1,3 +1,4 @@
+import contextlib
 import os
 import shutil
 import uuid as uuid_pkg
@@ -406,8 +407,12 @@ def preprocess_data(db: Session, file_id: uuid_pkg.UUID, transformations: list) 
         # Write to a temp file, then atomically replace the original so an
         # interruption between writing and replacing never leaves a truncated CSV.
         tmp_path = file_path + ".tmp"
-        df.to_csv(tmp_path, index=False)
-        os.replace(tmp_path, file_path)
+        try:
+            df.to_csv(tmp_path, index=False)
+            os.replace(tmp_path, file_path)
+        finally:
+            with contextlib.suppress(OSError):
+                os.remove(tmp_path)
         return _resp(200, True, "Dataset preprocessed successfully")
     except ValueError as e:
         return _resp(422, False, str(e))
--- a/tensormap-backend/app/services/data_process.py
+++ b/tensormap-backend/app/services/data_process.py
@@ -1,3 +1,5 @@
+import os
+import shutil
 import uuid as uuid_pkg
 from collections.abc import Callable
 from typing import Any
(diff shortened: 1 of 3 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `tensormap-backend/app/services/data_process.py:400` The .bak file is written after all transformations are applied to df, not before loading the file. The backup therefore captures the original on-disk file (pre-transform), which is correct, but the backup is created immediately before the overwrite with no atomic rename or temp-file pattern. If the process is interrupted between shutil.copy2 and df.to_csv, the backup exists but the CSV is partially written and truncated, leaving the data in an inconsistent state. Consider writing to a temp file first and then doing an os.replace for atomicity.
- **[minor/quality]** `tensormap-backend/app/services/data_process.py:400` Accumulated .bak files are never cleaned up. Every call to preprocess_data silently overwrites the previous .bak, so only one backup is retained, but there is no documentation or comment noting this limitation. If users run preprocessing multiple times (e.g. correcting a mistake), earlier backup is silently lost without warning.
- **[minor/quality]** `tensormap-backend/app/services/data_process.py:402` The except clause catches OSError only, but shutil.copy2 can also raise a PermissionError (which is a subclass of OSError on Python 3 so it is covered) and SameFileError (also OSError). The handling is therefore correct; however, after logging the warning the function continues and overwrites the original with no backup silently. The log level should arguably be ERROR, not WARNING, because the user has no way to recover their original data if the backup could not be written.
- **[nit/test]** `tensormap-backend/app/services/data_process.py:398` No test covers the backup path: neither the happy path (backup file created at file_path + '.bak') nor the failure path (OSError during copy, warning logged, csv still written). Given that the entire PR is this one behavior, at least one pytest case should be added.

**Conventions**

The PR adds a best-effort shutil.copy2 backup before overwriting the CSV, which addresses a real data-loss risk. The core logic is sound, but the write is not atomic: an interruption between the copy and df.to_csv leaves the file truncated with the backup already in place, so recovery is unreliable. Downgrading the warning to a silent continue when the backup fails is also risky because the user has no indication their original is unprotected. The change needs at least one pytest case covering both the happy and failure paths before merging.

ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `tensormap-backend/tests/test_data_process_backup.py:27` test_backup_created_on_happy_path does not patch pd.read_csv or select, so preprocess_data will try to query the real DB and read a real file before ever reaching the backup logic. The test will likely raise an AttributeError or TypeError on the mock_db.exec chain before the assertion is reached, making the test unreliable as a regression guard. _(unverified)_
- **[minor/bug]** `tensormap-backend/app/services/data_process.py:406` If df.to_csv(tmp_path) raises (e.g. disk full), the .tmp file is left on disk in the upload folder and is never cleaned up. A finally block that removes tmp_path on failure would prevent stale temporaries from accumulating.
- **[minor/security]** `tensormap-backend/app/services/data_process.py:399` The .bak file is written into the same upload folder as the original CSV with no access controls, and is exposed at a guessable path (original name + .bak). If the file-serving routes do not restrict access by file ID or session, any user who knows the original file name can download the backup directly.
- **[nit/quality]** `tensormap-backend/app/services/data_process.py:400` The comment 'Only one .bak is retained -- repeated preprocessing silently replaces it' describes a deliberate limitation that is never surfaced to the caller. If repeated preprocessing is a supported workflow, the user has no way to know their prior backup was overwritten. Consider returning a warning in the response body, or document this limitation in the API docs.

**Conventions**
- **Description:** The description explains what changed but does not link to a related issue or explain why this is needed now (no 'Closes #X' reference), making it harder to trace context.
- _Add a 'Closes #NNN' line to the PR description so the fix is traceable to the issue that motivated it._

The atomic write pattern (write to .tmp, then os.replace) is a solid improvement over the bare to_csv overwrite. However, the main test in the new test file does not patch pd.read_csv or select, so it will fail before exercising the backup logic it is meant to validate -- this needs to be fixed before the tests provide meaningful coverage. The .bak file is placed in the upload folder with no cleanup path for the orphaned .tmp on write failure, and with no access control consideration; both are worth addressing. The PR description should also link to the issue it closes.

PR #364, opened 2026-06-18
fix: replace server-local paths in generated training scripts with client-relative filenames

shivv23:
Downloaded training scripts contained hardcoded server-local paths (./data/uuid.csv, ./templates/json-model/model.json) that do not exist on the user's machine. Every generated script crashed on the first pd.read_csv() call.

Changed the file path to use the original uploaded filename (file.file_name) and the model JSON path to use just the model name, so both reference files the user can place alongside the script.

### DIFF
--- a/tensormap-backend/app/services/code_generation.py
+++ b/tensormap-backend/app/services/code_generation.py
@@ -1,7 +1,6 @@
 from jinja2 import Environment, FileSystemLoader
 from sqlmodel import Session, select
 
-from app.config import get_settings
 from app.models import DataFile, ImageProperties, ModelBasic
 from app.shared.constants import (
     BATCH_SIZE,
@@ -14,7 +13,6 @@ from app.shared.constants import (
     LABEL_MODE,
     LEARNING_MODEL,
     MODEL_EPOCHS,
-    MODEL_GENERATION_LOCATION,
     MODEL_GENERATION_TYPE,
     MODEL_JSON_FILE,
     MODEL_METRIC,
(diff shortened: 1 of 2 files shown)


ivantha:
### Automated review summary

**Findings**
- **[minor/bug]** `tensormap-backend/app/services/code_generation.py:99` The ZIP branch strips the last dot-segment from file.file_name (e.g. 'images.zip' → 'images'), which is the intended behaviour. However, if a user uploads a ZIP whose stored file_name lacks a '.' (e.g. 'images'), rsplit('.', 1)[0] silently returns the full string unchanged — the generated path would then be 'images' pointing at the wrong location. A guard or assertion that file_name ends with '.zip' would make the assumption explicit and catch data inconsistencies early.
- **[nit/quality]** `tensormap-backend/tests/test_null_checks.py:91` _file_location no longer has a null-safety concern (the caller now guards against file=None before calling it), but the function's type annotation says 'file: DataFile' with no Optional. This is correct and good; the removed test for 'raises_error_when_file_not_found' inside TestCodeGenFileLocation is adequately replaced by TestGenerateCode.test_raises_error_when_file_not_found. No gap in coverage, just confirming the migration is complete.

**Conventions**

This PR fixes a real crash: generated training scripts embedded server-local upload paths that do not exist on the user's machine, causing every pd.read_csv() and model-load call to fail immediately. The fix is correct — the null guard for the missing-file case was moved up into generate_code (which already fetched the DataFile object) and _file_location was simplified to a pure, testable function that takes the already-resolved object. The only minor concern is the ZIP extension-stripping logic, which silently no-ops when file_name contains no dot; a cheap assertion or explicit .endswith('.zip') check would make that assumption explicit and protect against data inconsistencies. Tests were updated in line with the refactor and coverage is maintained. Approve once the ZIP edge case is considered.

PR #365, opened 2026-06-18
fix: add value prop to Dataset File, Problem Type and Metric Select components

shivv23:
Three shadcn/ui Select components (Dataset File, Problem Type, Metric) were missing the value prop, making them uncontrolled. After config is saved and the user navigates away then back, the dropdowns show the placeholder instead of the saved value.

Added value={trainingConfig.X || ''} to each Select so they reflect the current state on re-render.


ivantha:
### Automated review summary

**Findings**
- None flagged.

**Conventions**

The fix is correct: all three shadcn/ui Select components were uncontrolled (missing the `value` prop), so navigating away and back caused the dropdowns to revert to their placeholder despite saved config state. Adding `value={trainingConfig.X || ""}` to each one aligns them with the rest of the form. The diff is minimal and surgically scoped. PR title, description, and single-commit history are all clean — this is ready to merge.

PR #368, opened 2026-06-22
fix: prevent duplicate DataProcess records by adding unique constraint and upsert

shivv23:
## Problem

`DataProcess.file_id` has no unique constraint. `add_target_service` always INSERTs. Calling it twice for the same file creates duplicate rows. When training queries the target with `.first()`, it picks an arbitrary duplicate — users unknowingly train against the wrong column.

## Root Cause

`app/services/data_process.py:54` — always creates a new `DataProcess(...)` without checking for an existing record for that `file_id`.

## Solution

Three-layer fix:

1. **DB model** (`app/models/data.py:61-63`) — `UniqueConstraint("file_id", name="uq_data_process_file_id")` on `DataProcess` so the database enforces one-target-per-file at the storage level.

2. **Service** (`app/services/data_process.py:58-63`) — `add_target_service` now queries for an existing `DataProcess` by `file_id` before inserting; updates `target` in place if found, creates a new record otherwise.

3. **Migration** (`migrations/versions/6a7b8c9d0e1f_...`) — before adding the constraint, deduplicates any existing rows by keeping the most recently updated row per `file_id`; drops the now-redundant non-unique index `ix_data_process_file_id`.

## Risk Assessment

| Concern | Mitigation |
|---------|------------|
| Migration crashes on existing duplicates | Dedup SQL runs before constraint — keeps row with latest `updated_on`, removes rest |
| Race: two concurrent POSTs both INSERT | UniqueConstraint catches the loser; `except Exception` returns 500 + message |
| `onupdate=func.now()` doesn't fire | Verified: SQLAlchemy fires `onupdate` on any UPDATE flush — automatic timestamp |
| Dropped index slows file_id queries | `UniqueConstraint` creates its own btree index — zero performance loss |

## Testing

- All 94 existing data-process-related tests pass unchanged.
- New test `test_updates_existing_target` verifies the upsert path (in-place update, `add` not called).
- Existing `test_success` renamed to `test_creates_new_target` — now mocks two `exec()` calls (file lookup + existing-target check).
- All other existing tests adjusted to the two-`exec()` pattern.

### DIFF
--- a/tensormap-backend/app/models/data.py
+++ b/tensormap-backend/app/models/data.py
@@ -2,7 +2,7 @@ import uuid as uuid_pkg
 from datetime import datetime
 from typing import TYPE_CHECKING, Optional
 
-from sqlalchemy import JSON, CheckConstraint, Column, DateTime, ForeignKey, func
+from sqlalchemy import JSON, CheckConstraint, Column, DateTime, ForeignKey, UniqueConstraint, func
 from sqlalchemy.dialects.postgresql import UUID as PgUUID
 from sqlmodel import Field, Relationship, SQLModel
 
@@ -57,6 +57,8 @@ class DataProcess(SQLModel, table=True):
     )
 
     file: DataFile | None = Relationship(back_populates="target")
+
+    __table_args__ = (UniqueConstraint("file_id", name="uq_data_process_file_id"),)
 
 
 class ImageProperties(SQLModel, table=True):
--- a/tensormap-backend/app/services/data_process.py
+++ b/tensormap-backend/app/services/data_process.py
@@ -41,7 +41,11 @@ def _get_file_path(file: DataFile) -> str:
 
 
 def add_target_service(db: Session, file_id: uuid_pkg.UUID, target: str) -> tuple:
-    """Create a DataProcess record linking a file to its target column."""
+    """Create or update the target field assignment for a file.
+
+    Upserts: if a DataProcess record already exists for this file_id the
+    target column is updated; otherwise a new record is created.
+    """
     try:
         file = db.exec(select(DataFile).where(DataFile.id == file_id)).first()
         if not file:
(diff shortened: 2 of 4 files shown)


ivantha:
### Automated review summary

**Findings**
- **[minor/bug]** `tensormap-backend/app/services/data_process.py:67` HTTP 201 (Created) is returned on the update path as well as the create path. REST convention says updates should return 200 (OK). Both branches fall through to the same `return _resp(201, True, 'Target field added successfully')` at line 67, so a client that inspects the status code cannot tell whether a record was created or updated.
- **[minor/quality]** `tensormap-backend/app/services/data_process.py:60` The upsert has a TOCTOU race: two concurrent POST requests can both reach the `existing = None` branch and both attempt INSERT. The unique constraint will reject the second attempt, but the bare `except Exception` handler (inherited from the surrounding try/except) will surface that as a generic 500. Consider catching `sqlalchemy.exc.IntegrityError` explicitly and retrying or returning a 409 Conflict, to give the caller a meaningful response.
- **[nit/quality]** `tensormap-backend/migrations/versions/6a7b8c9d0e1f_add_unique_constraint_on_data_process_file_id.py:7` The migration revision ID `6a7b8c9d0e1f` is a hand-crafted sequential hex string rather than the random UUID Alembic generates. This is not wrong, but it increases the chance of a collision if another migration is generated around the same time. Prefer Alembic-generated IDs (`alembic revision --autogenerate`) for consistency with the rest of the project.
- **[nit/test]** `tensormap-backend/tests/test_data_process_service.py:107` The log assertion string in `test_creates_new_target` is not updated after the rename: the test verifies `body['message'] == 'Target field added successfully'` but the test itself is now exercising the create-only path. Adding a parallel assertion for the update path in `test_updates_existing_target` (checking the same message) would confirm message consistency across both branches.

**Conventions**

This PR correctly addresses a real data-integrity bug: `add_target_service` previously always INSERTed, allowing duplicate `DataProcess` rows per file. The three-layer fix (unique constraint on the model, upsert in the service, deduplication migration) is architecturally sound and the migration handles pre-existing duplicates safely. The main actionable issue is that the upsert returns HTTP 201 on both create and update paths -- the update branch should return 200 so callers can distinguish the two outcomes. The TOCTOU race under concurrent requests is low-risk but worth hardening with a specific `IntegrityError` handler rather than relying on the generic `except Exception`. The migration revision ID should be Alembic-generated rather than hand-crafted to avoid future collisions.

PR #369, opened 2026-06-25
fix: show real backend error messages instead of '[object Object]' on training failure

shivv23:
## Problem

When training fails, users see **"[object Object]"** instead of the actual backend error message. This makes debugging training failures impossible — the user (and developer) has zero actionable information.

## Root Cause

Two broken links in the error propagation chain:

**Link 1 — `ModelServices.js:189`:**
```js
throw new Error(resp.data);
```
`resp.data` is an object `{success: false, message: "GPU OOM", data: null}`. JavaScript calls `.toString()` on it, producing `Error('[object Object]')`.

*Note: currently the training endpoint returns HTTP 400/404 for errors, so this `.then()` line is dead code for this specific endpoint — axios rejects before reaching it. But it's a bug waiting to happen if the API design ever changes.*

**Link 2 — `Training.jsx:487-492`:**
```js
logger.error(error.response?.data);
setResultValues([error.response?.data?.message ?? "An error occurred"]);
```
Only reads `error.response?.data?.message`. This works for axios errors (HTTP 400 with our standard `{message: "..."}` response), but fails silently for:
- Plain `Error` objects thrown from `runModel`'s `.then()` (no `.response` property)
- FastAPI HTTP 500 responses (`{detail: "..."}` — no `.message` field)
- Network errors (no `.response` at all)

In all those cases the fallback `"An error occurred"` is shown, losing the real error.

## Solution

**`ModelServices.js:189`** — Extract the message string instead of passing the raw object:
```js
// Before:
throw new Error(resp.data);
// After:
throw new Error(resp.data?.message ?? JSON.stringify(resp.data));
```

**`Training.jsx:487-492`** — Broaden the catch handler to extract messages from all error shapes, and log the full error object for debugging:
```js
logger.error(error);   // full error with stack trace
const message =
  error.response?.data?.message ??                              // backend standard format
  (Array.isArray(error.response?.data?.detail)                   // FastAPI 422 array
    ? error.response.data.detail.map((d) => d.msg).join("; ")
    : error.response?.data?.detail) ??                          // FastAPI 500 string
  error.message ??                                               // plain Error / network error
  "An error occurred";
setResultValues([message]);
```

### Error Decoding Priority

| Source | Example | Reads |
|--------|---------|-------|
| Backend standard response | `{success: false, message: "Model not found"}` | `error.response.data.message` |
| FastAPI 500 | `{detail: "Internal server error"}` | `error.response.data.detail` |
| FastAPI 422 (validation) | `{detail: [{msg: "field required"}]}` | `error.response.data.detail[0].msg` |
| Plain Error / Network | `Error("timeout of 300s exceeded")` | `error.message` |
| Everything else | — | `"An error occurred"` |

## Verification

- **Lint:** `npm run lint` exits clean (0 errors, 0 warnings)
- **Tests:** All 80 frontend tests pass across 15 test files
- **No new dependencies added**

## Risk Assessment

| Concern | Mitigation |
|---------|------------|
| `error.response?.data?.detail` could be an array (FastAPI 422) without `.msg` | Protected by `Array.isArray` check + `.map((d) => d.msg)` — if no `.msg`, produces `""` and falls through to next `??` |
| Logger change loses stack trace | Reversed: old code logged `error.response?.data` (only response body, no stack). New code logs `error` (full Error object with stack trace) |
| Fixing dead code that never executes | Defensive — if endpoint design changes to use 200+`success:false`, this prevents `[object Object]` from appearing |
| Two files must ship together | Single atomic commit — no partial-deployment risk |

## Files Changed

| File | Change |
|------|--------|
| `tensormap-frontend/src/services/ModelServices.js:189` | Extract `resp.data?.message` instead of raw object |
| `tensormap-frontend/src/containers/Training/Training.jsx:487-492` | Multi-source error message extraction + full error logging |

### DIFF
--- a/tensormap-frontend/src/containers/Training/Training.jsx
+++ b/tensormap-frontend/src/containers/Training/Training.jsx
@@ -486,8 +486,15 @@ export default function Training() {
       .then(() => {})
       .catch((error) => {
         clearTimeout(timeoutRef.current);
-        logger.error(error.response?.data);
-        setResultValues([error.response?.data?.message ?? "An error occurred"]);
+        logger.error(error);
+        const message =
+          error.response?.data?.message ??
+          (Array.isArray(error.response?.data?.detail)
+            ? error.response.data.detail.map((d) => d.msg).join("; ")
+            : error.response?.data?.detail) ??
+          error.message ??
+          "An error occurred";
+        setResultValues([message]);
         setIsLoading(false);
       });
   }, [selectedModel, projectId, validateAllFields]);
--- a/tensormap-frontend/src/services/ModelServices.js
+++ b/tensormap-frontend/src/services/ModelServices.js
@@ -186,7 +186,7 @@ export const runModel = async (modelName, projectId) => {
       if (resp.data.success === true) {
         return resp.data.message;
       }
-      throw new Error(resp.data);
+      throw new Error(resp.data?.message ?? JSON.stringify(resp.data));
     })
     .catch((err) => {
       throw err;


ivantha:
### Automated review summary

**Findings**
- **[minor/quality]** `tensormap-frontend/src/containers/Training/Training.jsx:493` The FastAPI 422 array branch calls `detail.map((d) => d.msg).join("; ")`. JavaScript's `Array.prototype.join()` converts `undefined` to an empty string, so if any detail item lacks a `msg` field the result is a string of semicolons (e.g. `"; "`) rather than an empty/falsy value. The `??` operator therefore never falls through to `error.message`, and the user sees a confusing near-empty string instead of a network-level error message.
- **[minor/test]** `tensormap-frontend/src/containers/Training/Training.jsx:490` The new catch handler has four distinct branches (standard `message`, FastAPI 422 detail array, FastAPI detail string, plain Error / network error). No new tests cover any of them. These branches are straightforward to exercise with a mocked Axios error and a Jest test, and the absence of coverage means a future regression in error extraction would go undetected.

**Conventions**

This is a focused, well-diagnosed bug fix that corrects two separate breakpoints in the error propagation chain: the raw-object coercion in ModelServices.js and the narrow message extraction in Training.jsx. The PR description is exemplary, tracing root causes, enumerating all error shapes, and providing a risk table. One minor quality concern: the FastAPI 422 branch's `map((d) => d.msg).join("; ")` produces a truthy string of semicolons rather than falling through when detail items lack a `msg` field, so a guard such as `.filter(Boolean)` or a fallback after the join is needed. No new tests cover the four new error-handling branches; adding even a lightweight Jest test for each shape would make the logic regression-proof.

PR #371, opened 2026-06-27
feat(training): Training Job Persistence, MetricsCallback, and Socket.IO Room Isolation

kartikeyg0104:
## Summary

Replaces TensorMap's stateless, fire-and-block training flow with **persistent, cancellable, per-user-isolated training jobs**.

Previously, `POST /model/run` ran training synchronously inside the request (the HTTP call hung for the entire run), streamed progress as **formatted text strings**, and **broadcast them to every connected client**. There was no record of past runs, no way to cancel, and two users training at once saw each other's progress.

Now:
- Two new tables (`training_job`, `training_metric`), a job lifecycle state machine, structured per-epoch metric persistence via a new `MetricsCallback`, cooperative cancellation via `CancellationCheckCallback`, crash recovery on startup, and a concurrency guard.
- Per-job Socket.IO **rooms** (metrics emitted to `room=job_id`,  never broadcast), a`subscribe_job`handler with **catch-up replay** for late/reconnecting clients, and a frontend **Stop Training** button.

## Architecture

```
POST /model/run ──> create TrainingJob (PENDING) ──> 202 {job_id}  (returns immediately)
                          │
                          └─ background thread ─> model_run(job_id)
                                                   ├─ MetricsCallback ─> training_metric rows + emit room=job_id
                                                   └─ CancellationCheckCallback ─> stops at epoch boundary if CANCELLED

client ──subscribe_job(job_id)──> joins room ──> catch-up replay + live metrics
       ──DELETE /training-job/:id──> status=CANCELLED ──> training stops, terminal status emitted
```

Job lifecycle: `PENDING → RUNNING → COMPLETED` (happy path), or `FAILED` (exception) / `CANCELLED` (user stop). Jobs left active by a crash are marked `FAILED` on startup (`orphan_recovery`).

## Endpoints (all under `/api/v1/model`)

| Method | Path | Notes |
|--------|------|-------|
| `POST` | `/run` | Creates a job, fires it in the background, returns **202** with `job_id` |
| `GET` | `/training-job/{job_id}` | Status, hyperparams, latest metrics |
| `GET` | `/training-job/{job_id}/metrics` | Full per-epoch history (chart-ready) |
| `DELETE` | `/training-job/{job_id}` | Requests cancellation (**204**) |
| `GET` | `/training-jobs?model_name=…` | Jobs for a model, newest first |

## Deliberate deviations from the original spec

The spec assumed infrastructure that does not exist in this codebase. These were adapted as senior-dev calls and documented inline:

1. **No authentication / `User` model exists.** Jobs are scoped to the existing`project_id` instead of `owner_id`. `verify_job_owner` became `get_job_or_404` (returns **404** for unknown jobs; there is no per-user authorization to enforce, so no 403).
2. **Synchronous DB** (`sqlmodel.Session`, not `AsyncSession`). Sync DB reads inside async Socket.IO handlers run via `asyncio.to_thread`.
3. **`model_basic.id` is an `int`**, so `TrainingJob.model_id` is `int`.
4. **`_main_loop` global was already removed** (refactored to a `ContextVar` in a prior PR) — no change needed there.
5. **Two divergent Alembic heads existed** (`alembic upgrade head` was ambiguous). Added a **merge migration** to unify them before the training-tables migration — fixes a latent bug.
6. **Response envelope.** All endpoints (including errors, via `AppException`) follow the repo's documented `{success, message, data}` format.

## Concurrency & safety details

- The run endpoint returns **202 immediately**; training runs in a tracked background task (`asyncio.to_thread`) with its **own** DB session — a strong reference is held so the task isn't garbage-collected mid-run.
- `on_train_end` only flips `RUNNING → COMPLETED`, so a `CANCELLED`/`FAILED` job is never overwritten with `COMPLETED`.
- The concurrency guard counts `PENDING + RUNNING` (not just `RUNNING`) to avoid a burst of quick requests slipping past before any flips to `RUNNING`.
- Status enum stored as a portable `VARCHAR + CHECK` (`native_enum=False`, lowercase `values_callable`) — no Postgres ENUM type to migrate.

## Frontend

- `services/socketService.js` — shared Socket.IO client with `subscribeToJob`/`unsubscribeFromJob`/`cancelJob`. **Re-subscribes on every (re)connect** so a reconnect re-joins the room and triggers catch-up (no silently-missed metrics).
- `Training.jsx` — subscribes to the started job, renders structured metrics, and shows a **Stop Training** button (disabled with "Cancellation requested…" after click). Charts are intentionally deferred to Week 6.

### DIFF
--- a/tensormap-backend/app/callbacks/__init__.py
+++ b/tensormap-backend/app/callbacks/__init__.py
@@ -1,0 +1,6 @@
+"""Keras callbacks for persisting and streaming training progress."""
+
+from app.callbacks.cancellation_callback import CancellationCheckCallback
+from app.callbacks.metrics_callback import MetricsCallback
+
+__all__ = ["CancellationCheckCallback", "MetricsCallback"]
--- a/tensormap-backend/app/callbacks/cancellation_callback.py
+++ b/tensormap-backend/app/callbacks/cancellation_callback.py
@@ -1,0 +1,38 @@
+"""Keras callback that cooperatively cancels training when requested.
+
+The DELETE endpoint only sets the job's status to CANCELLED in the DB; this
+callback is what actually stops the Keras loop. It checks the status at each
+epoch boundary and sets ``model.stop_training`` so ``fit`` exits cleanly.
+"""
+
+import tensorflow as tf
+
+from app.models.training_job import TrainingJob, TrainingStatus
+from app.shared.logging_config import get_logger
+
+logger = get_logger(__name__)
+
+
+class CancellationCheckCallback(tf.keras.callbacks.Callback):
+    """Stop training when the job's DB status becomes CANCELLED.
+
+    Args:
+        job_id: The training job to watch.
+        session_factory: Zero-arg callable returning a fresh DB ``Session``.
+    """
+
+    def __init__(self, job_id, session_factory) -> None:
+        super().__init__()
+        self.job_id = job_id
+        self.session_factory = session_factory
+        # Side-effect flag so tests (and callers) can confirm a cancellation fired.
+        self.cancelled = False
+
+    def on_epoch_begin(self, epoch: int, logs: dict = None) -> None:
+        """Set ``model.stop_training`` if the job has been cancelled."""
+        with self.session_factory() as session:
+            job = session.get(TrainingJob, self.job_id)
+            if job is not None and job.status == TrainingStatus.CANCELLED:
+                logger.info("Cancellation detected for job %s at epoch %d", self.job_id, epoch)
+                self.cancelled = True
+                self.model.stop_training = True
--- a/tensormap-backend/app/callbacks/metrics_callback.py
+++ b/tensormap-backend/app/callbacks/metrics_callback.py
@@ -1,0 +1,136 @@
+"""Keras callback that persists per-epoch metrics and streams them to a job room.
+
+Replaces the old ``CustomProgressBar`` text emissions: instead of broadcasting
+formatted strings to every client, it writes structured metrics to the
+``training_metric`` table and emits them to the Socket.IO room for this job only.
+"""
+
+import asyncio
+from datetime import UTC, datetime
+
+import tensorflow as tf
+
+from app.models.training_job import TrainingJob, TrainingStatus
+from app.models.training_metric import TrainingMetric
+from app.shared.constants import SOCKETIO_DL_NAMESPACE, SOCKETIO_LISTENER
+from app.shared.logging_config import get_logger
+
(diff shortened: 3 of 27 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `tensormap-backend/app/routers/training.py:91` Hyperparameter overrides accepted by POST /run (epochs, batch_size, optimizer, lr) are stored in job.hyperparams but never forwarded to model_run. launch_training_job passes only model_name and job_id; _train_in_thread calls model_run without any hyperparams context, so model_run reads epochs/batch_size/etc. directly from ModelBasic in the DB and the request overrides are silently ignored. The endpoint docstring says they 'override [stored values] for this run', making the API contract misleading.
- **[minor/quality]** `tensormap-backend/app/callbacks/metrics_callback.py:40` schedule_room_emit calls future.result(timeout=5), blocking the Keras training thread for up to 5 seconds on every per-epoch metric emit. On fast-epoch models (GPU, small datasets) this throttles throughput proportionally to epoch count. Making the emit fire-and-forget — scheduling the coroutine and logging a warning if the future is not done within a brief poll — would remove this blocking from the hot training path.
- **[nit/quality]** `tensormap-backend/app/routers/training.py:111` list_training_jobs orders by started_at.desc(), but started_at is NULL for PENDING jobs (it is set by on_train_begin, not at creation). PostgreSQL places NULLs last in DESC order by default, so a freshly created PENDING job appears below all completed/failed runs in the list response, which is counterintuitive.
- **[nit/quality]** `tensormap-backend/app/models/training_job.py:63` Three columns reserved for later phases (analysis_cache, last_export_download_at, tuning_session_id) are shipped unused in the training_job table. Adding nullable columns speculatively couples the current migration to designs that have not been finalised; if those phases need different types or constraints, a new migration will be required regardless.

**Conventions**

This PR delivers a well-structured persistence and isolation layer for training jobs, replacing the old fire-and-block flow with cancellable DB-backed jobs, per-job Socket.IO rooms, startup orphan recovery, and broad test coverage including end-to-end Keras runs with real callbacks. The primary issue to resolve before merging: hyperparameter overrides (epochs, batch_size, lr, optimizer) accepted by POST /run are recorded in job.hyperparams but not forwarded to model_run, so actual training always uses the model's saved DB config and the request overrides are silently ignored despite the endpoint docstring claiming they apply for this run. A secondary concern is that schedule_room_emit uses future.result(timeout=5), blocking the Keras training thread up to 5 seconds on every epoch emit; consider scheduling the coroutine fire-and-forget and logging a warning on failure instead.

PR #374, opened 2026-06-28
fix: seed migration uses correct CSV filename and encode string labels before training

### LINKED ISSUE #373: [Bug] Iris Sample Project training fails: seed migration missing disk_name and wrong source CSV filename
## Describe the bug
The pre-seeded "Iris Sample Project" fails to train with "Model running failed. 
Please recheck the model configs" error.

## Root Cause
Two issues in `f4a8c2e91b37_seed_sample_project_pipeline.py`:

**Issue 1:** The seed migration copies `test_dataset.csv` → `iris_sample.csv` but 
the container only ships `sample_data.csv`. The copy never executes, so 
`iris_sample.csv` never exists on disk.

**Issue 2:** The `data_file` INSERT does not set the `disk_name` column 
(added in migration `e7b9a4c20e11`). So `disk_name` is NULL, and 
`_helper_generate_file_location` resolves the path as `./data/None`, 
causing a `FileNotFoundError` when training starts.

## Steps to Reproduce
1. Fresh `docker-compose up --build`
2. Open Iris Sample Project → Training
3. Fill in config and click Train
4. Observe: "Model running failed. Please recheck the model configs"

## Backend error
FileNotFoundError: [Errno 2] No such file or directory: './data/iris_sample.csv'

Image

## Expected behavior
Training should start successfully on the pre-seeded Iris dataset.

## Fix
1. Change `SOURCE_CSV` to use `sample_data.csv` (the file that actually exists)
2. Add `disk_name` to the `data_file` INSERT in the seed migration

ashishsoni-ai:
## What
Fixes #373 — Iris Sample Project training fails out of the box on a fresh install.

## Root Cause
Two bugs found through local debugging:

**Bug 1:** `f4a8c2e91b37_seed_sample_project_pipeline.py` references `test_dataset.csv` as the source CSV but the container ships `sample_data.csv`. The copy never executes so `iris_sample.csv` is never created on disk, causing `FileNotFoundError` at training time.

**Bug 2:** `_prepare_training_data` in `model_run.py` did not encode string target labels (e.g. `setosa`, `versicolor`, `virginica`) to integers before passing to `model.fit`, causing `ValueError: Invalid dtype: str`.

## Fix
1. Change `SOURCE_CSV` to `sample_data.csv` in the seed migration
2. Add `pd.Categorical(y).codes` encoding for non-numeric target columns

## Tested
Fresh `docker-compose up --build` → Iris Sample Project → Train → training completes successfully for all 10 epochs.

### DIFF
--- a/tensormap-backend/app/services/model_run.py
+++ b/tensormap-backend/app/services/model_run.py
@@ -229,6 +229,9 @@ def _prepare_training_data(features, target_field, training_split):
 
     X = features.drop(target_field, axis=1)
     y = features[target_field]
+    # Encode string labels to integers for categorical targets
+    if not pd.api.types.is_numeric_dtype(y):
+        y = pd.Categorical(y).codes
 
     split_index = int(len(X) * float(training_split) / 100)
     return X[:split_index], y[:split_index], X[split_index:], y[split_index:]
--- a/tensormap-backend/migrations/versions/f4a8c2e91b37_seed_sample_project_pipeline.py
+++ b/tensormap-backend/migrations/versions/f4a8c2e91b37_seed_sample_project_pipeline.py
@@ -29,7 +29,7 @@ FILE_UUID = "00000000-0000-4000-a000-000000000002"
 # File paths (relative to tensormap-backend/, where alembic runs)
 DATA_DIR = "./data"
 MODEL_DIR = "./templates/json-model"
-SOURCE_CSV = os.path.join(DATA_DIR, "test_dataset.csv")
+SOURCE_CSV = os.path.join(DATA_DIR, "sample_data.csv")
 TARGET_CSV = os.path.join(DATA_DIR, "iris_sample.csv")
 MODEL_JSON_PATH = os.path.join(MODEL_DIR, "iris-classifier.json")
 


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `tensormap-backend/app/services/model_run.py:193` pd.Categorical(y).codes returns numpy.int8. TensorFlow/Keras label tensors typically require int32 or int64; passing int8 can cause silent dtype mismatches or explicit cast errors depending on the Keras version. Add .astype('int64') (or at minimum 'int32') after .codes.
- **[major/bug]** `tensormap-backend/app/services/model_run.py:193` pd.Categorical(y).codes returns -1 for NaN entries in y. If the target column has any missing values, -1 codes are silently forwarded to model.fit as label indices, which will either produce wrong results or a cryptic out-of-range index error. A NaN check and early ValueError should precede the encoding.
- **[minor/quality]** `tensormap-backend/app/services/model_run.py:191` The category-to-integer mapping is silently discarded. Downstream callers and any future prediction or evaluation code have no way to map predicted class indices back to the original string labels. Consider returning or logging the mapping alongside the encoded y.
- **[minor/test]** `tensormap-backend/app/services/model_run.py:191` No test is added for the new non-numeric target encoding path. A unit test covering _prepare_training_data with a string-label column (including a NaN-containing variant) would prevent regressions.

**Conventions**
- **PR title:** Verb tense is inconsistent: 'uses correct CSV filename' (indicative) vs 'encode string labels' (bare infinitive); should read 'encodes ... and encodes ...' for parallel structure.
- _Description is thorough and links the issue clearly; only the PR title has a minor parallelism issue._

The PR addresses two genuine bugs that prevent the Iris sample project from training on a fresh install, and the direction of both fixes is correct. The migration filename change is clean. The label encoding fix needs two hardening changes before merge: cast `.codes` from int8 to int64 to avoid dtype issues with Keras, and guard against NaN values in the target column (currently they silently produce -1 codes, which are invalid label indices). No tests cover the new encoding path, which should be added given that this is a shared utility used by all training runs.

PR #375, opened 2026-06-28
fix: add missing required env vars to .env.example

ashishsoni-ai:
## What
Adds missing required environment variables to `tensormap-backend/.env.example`.

## Problem
`app/config.py` requires `DATABASE_URL` and `SECRET_KEY` with no defaults.
The old `.env.example` only had Postgres vars, so a fresh contributor running
the backend locally would immediately hit a `ValidationError` with no guidance
on what to set.

## Fix
Added all required and optional variables with comments explaining each:
- `DATABASE_URL` — includes a SQLite example for local dev without Postgres
- `SECRET_KEY` — with a note to change in production
- `CORS_ALLOWED_ORIGINS`
- `UPLOAD_FOLDER`
- `MAX_CONTENT_LENGTH`
- `DEBUG`

### DIFF
--- a/tensormap-backend/.env.example
+++ b/tensormap-backend/.env.example
@@ -1,7 +1,19 @@
-# For local dev, use localhost. In Docker, this is overridden by docker-compose.yml environment block.
-DATABASE_URL=postgresql+psycopg2://<db-user>:<db-password>@localhost:5432/<db-name>
-SECRET_KEY=<generate-a-secret-key>
-# Comma-separated list of allowed origins (e.g. http://localhost:3300,https://app.example.com)
+# PostgreSQL configuration (used by docker-compose db service)
+POSTGRES_DB=c2si_db
+POSTGRES_USER=c2si
+POSTGRES_PASSWORD=c2si
+
+# Application settings
+# For local dev without Docker, use sqlite:///./tensormap.db
+DATABASE_URL=postgresql://c2si:c2si@localhost:5432/c2si_db
+SECRET_KEY=change-me-in-production
+
+# CORS origins (comma-separated)
 CORS_ALLOWED_ORIGINS=http://localhost:3300
+
+# File upload settings
 UPLOAD_FOLDER=./data
+MAX_CONTENT_LENGTH=209715200
+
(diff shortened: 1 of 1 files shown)


ivantha:
### Automated review summary

**Triage:** docs  |  **Recommendation:** ready-for-maintainer

**Findings**
- **[minor/security]** `tensormap-backend/.env.example:9` SECRET_KEY=change-me-in-production looks like a valid string value rather than a clearly unusable placeholder. A contributor who copies the file and forgets to replace it will end up with a known, weak secret key used for session signing. A comment-only instruction or an angle-bracket placeholder (e.g. SECRET_KEY=<run: python -c "import secrets; print(secrets.token_hex(32))">) makes it unambiguous that replacement is required.
- **[nit/quality]** `tensormap-backend/.env.example:8` The original DATABASE_URL used the explicit postgresql+psycopg2:// dialect specifier; this PR changes it to postgresql://. Modern SQLAlchemy resolves this to psycopg2 automatically when only that adapter is installed, so it usually works, but the explicit form is more robust across SQLAlchemy versions and matches what the codebase historically documented.
- **[nit/style]** `tensormap-backend/.env.example:19` File is missing a trailing newline on the last line (DEBUG=true). POSIX-compliant text files end with a newline; its absence can cause issues with diff tools and some linters.

**Conventions**
- _No GitHub issue is referenced; if a backing issue exists, add 'Closes #N' to the description._

This PR expands tensormap-backend/.env.example from 4 to 13 variables, adding PostgreSQL-specific vars, UPLOAD_FOLDER, and MAX_CONTENT_LENGTH with inline documentation, which is a genuine developer-experience improvement. The PR description is clear, and the single commit is well-formed. Two nits worth fixing before merge: the SECRET_KEY placeholder reads as a real value and could be silently copied into non-development deployments, so switching it to an angle-bracket or instruction-style placeholder reduces that risk; and the file is missing a trailing newline. The DATABASE_URL dialect change from postgresql+psycopg2:// to postgresql:// is benign in practice but loses the explicit driver specifier from the original.

PR #376, opened 2026-06-28
fix: getCovMatrix silently returns undefined on success:false, crashing PreprocessData

shivv23:
## Description

`getCovMatrix` in `FileServices.js:63-74` silently returns `undefined` when `resp.data.success !== true` because the `.then()` callback has no `return` statement on the fallthrough path. The caller `PreprocessData.jsx:32-33` then accesses `response.data_types` on `undefined`, causing a `TypeError`.

## Changes

1. **`FileServices.js:67-70`** — Added `throw new Error(resp.data.message || "Failed to fetch data metrics")` on `success: false`, matching the existing pattern in `getCorrelationMatrix` (line 89).

2. **`PreprocessData.jsx:36`** — Changed `setFetchError("Failed to load column data")` to `setFetchError(e.message)`, so users see the real backend error message instead of a generic string.

## Verification

- Lint: clean (0 errors, 0 warnings)
- Tests: 15 files, 80 tests passed (no regressions)

### DIFF
--- a/tensormap-frontend/src/components/Process/PreprocessData.jsx
+++ b/tensormap-frontend/src/components/Process/PreprocessData.jsx
@@ -33,7 +33,7 @@ function PreprocessData({ fileId, updateFileList }) {
         setData(Object.keys(response.data_types));
       } catch (e) {
         logger.error(e);
-        setFetchError("Failed to load column data");
+        setFetchError(e.message);
         setData(null);
       }
     };
--- a/tensormap-frontend/src/services/FileServices.js
+++ b/tensormap-frontend/src/services/FileServices.js
@@ -67,6 +67,7 @@ export const getCovMatrix = async (file_id) =>
       if (resp.data.success === true) {
         return resp.data.data;
       }
+      throw new Error(resp.data.message || "Failed to fetch data metrics");
     })
     .catch((err) => {
       logger.error(err);


ivantha:
### Automated review summary

**Findings**
- **[nit/quality]** `tensormap-frontend/src/services/FileServices.js:73` The thrown error is logged twice per failure: first by logger.error(err) in getCovMatrix's .catch() handler before rethrowing, and again by logger.error(e) in PreprocessData.jsx's catch block. This is consistent with the getCorrelationMatrix pattern already in the codebase but produces duplicate entries in the log for every success:false response. Consider logging only at the call site or only in the service layer.

**Conventions**
- **Description:** Description does not include a 'Closes #xxx' or 'Fixes #xxx' reference; if a tracking issue exists it should be linked.
- _Add a 'Closes #xxx' line to the description if a tracking issue exists; otherwise hygiene is clean._

This is a correct, minimal bug fix: getCovMatrix now throws an Error on success:false instead of silently falling through and returning undefined, which directly matches the existing pattern in getCorrelationMatrix. The .catch() handler rethrows, so the error propagates correctly to PreprocessData's try/catch, and setFetchError(e.message) surfaces the real backend message rather than a static string. The one nit is that the shared pattern causes the error to be logged twice (service layer + component), but that is pre-existing and consistent across the codebase. Fix is clean, tests pass, and is ready to merge after adding a Closes reference if a tracking issue exists.

PR #377, opened 2026-06-30
fix: CreateProjectDialog shows no error feedback on API failure

shivv23:
## Description

When the project creation API returns an error (backend 500, validation 422, network failure), `CreateProjectDialog` silently stops the loading spinner with no visible feedback — no error message, no toast, nothing. Users click "Create" repeatedly, unaware of what went wrong.

## Root Cause

`handleSubmit` has no `else` branch on `if (resp.data.success)` (line 41) and the `catch` block (line 47-49) only calls `logger.error`. No `errorMessage` state exists in the component.

## Changes

1. **`errorMessage` state** — Added `const [errorMessage, setErrorMessage] = useState("")` to track the error string.

2. **`else` branch** — On `resp.data.success === false`, sets error from `resp.data.message`. Defensive: backend uses non-2xx for errors today, but future changes could produce 2xx + `success: false`.

3. **`catch` block error extraction** — Extracts from three error shapes in priority order:
   - `err.response?.data?.message` (our API convention)
   - FastAPI `detail` (array → `.msg` join, or plain string)
   - `err.message` (network errors, JS errors)
   - Fallback: `"Failed to create project"`

4. **Error display** — `<div className="mb-4 text-sm text-destructive">` rendered before `DialogFooter`, shown conditionally when `errorMessage` is non-empty.

5. **Error reset** — `setErrorMessage("")` at the top of `handleSubmit` clears stale errors before each attempt.

## Verification

- Lint: clean (0 errors, 0 warnings)
- Tests: 15 files, 80 tests passed (no regressions)

### DIFF
--- a/tensormap-frontend/src/containers/ProjectsPage/CreateProjectDialog.jsx
+++ b/tensormap-frontend/src/containers/ProjectsPage/CreateProjectDialog.jsx
@@ -27,12 +27,14 @@ export default function CreateProjectDialog({ open, onOpenChange, onCreated }) {
   const [name, setName] = useState("");
   const [description, setDescription] = useState("");
   const [loading, setLoading] = useState(false);
+  const [errorMessage, setErrorMessage] = useState("");
 
   const handleSubmit = async (e) => {
     e.preventDefault();
     if (!name.trim()) return;
 
     setLoading(true);
+    setErrorMessage("");
     try {
       const resp = await createProject({
         name: name.trim(),
@@ -43,16 +45,34 @@ export default function CreateProjectDialog({ open, onOpenChange, onCreated }) {
         setName("");
         setDescription("");
         onOpenChange(false);
+      } else {
+        setErrorMessage(resp.data.message || "Failed to create project");
       }
     } catch (err) {
       logger.error("Failed to create project:", err);
+      const msg =
+        err.response?.data?.message ||
+        (Array.isArray(err.response?.data?.detail)
+          ? err.response.data.detail.map((d) => d.msg).join("; ")
+          : typeof err.response?.data?.detail === "string"
+            ? err.response.data.detail
+            : undefined) ||
+        err.message ||
+        "Failed to create project";
+      setErrorMessage(msg);
     } finally {
       setLoading(false);
     }
   };
 
   return (
-    <Dialog open={open} onOpenChange={onOpenChange}>
+    <Dialog
+      open={open}
+      onOpenChange={(open) => {
+        if (!open) setErrorMessage("");
+        onOpenChange(open);
+      }}
+    >
       <DialogContent>
         <DialogHeader>
           <DialogTitle>New Project</DialogTitle>
@@ -85,8 +105,20 @@ export default function CreateProjectDialog({ open, onOpenChange, onCreated }) {
               />
             </div>
           </div>
+          {errorMessage && (
+            <div className="mb-4 text-sm text-destructive" role="alert">
+              {errorMessage}
+            </div>
+          )}
           <DialogFooter>
-            <Button type="button" variant="outline" onClick={() => onOpenChange(false)}>
+            <Button
+              type="button"
+              variant="outline"
+              onClick={() => {
+                setErrorMessage("");
+                onOpenChange(false);
+              }}
+            >
               Cancel
             </Button>
             <Button type="submit" disabled={loading || !name.trim()}>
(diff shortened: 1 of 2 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `tensormap-frontend/src/containers/ProjectsPage/CreateProjectDialog.jsx:100` errorMessage is only cleared at the top of handleSubmit (line 37), never on Cancel or dialog dismissal, so a stale error banner from a previous failed attempt reappears the next time the dialog is opened.
- **[major/test]** `tensormap-frontend/src/containers/ProjectsPage/CreateProjectDialog.jsx:53` The new error-handling branches (success:false path at line 49 and the three-way catch-block message extraction at lines 53-60) ship with no accompanying unit test; the PR only claims the pre-existing 80 tests still pass, which doesn't exercise this new logic.
- **[minor/bug]** `tensormap-frontend/src/containers/ProjectsPage/CreateProjectDialog.jsx:57` When err.response.data.detail is neither an array nor a plain string (e.g. a backend raises HTTPException with a dict-shaped detail), the ternary falls through to that object itself, which is truthy, so a non-string value can reach setErrorMessage and later be rendered directly as a JSX child.
- **[minor/quality]** `tensormap-frontend/src/containers/ProjectsPage/CreateProjectDialog.jsx:100` The error banner is a plain div with no role="alert" or aria-live attribute, so screen-reader users are not notified when the error text appears.

**Conventions**
- **Description:** Root cause and change breakdown are thorough and accurate, but the description links no tracked issue (GitHub reports zero closing-issue references for this PR).
- _Otherwise clean hygiene (single well-scoped Conventional Commit matching the title); just add an issue reference if one exists._

The PR adds real value: it introduces error state and a visible banner covering the API success:false case and three shapes of thrown errors (API message, FastAPI validation detail, generic error), and its root-cause writeup accurately cites the pre-fix code. The main gap is that errorMessage is only reset when a new submit begins, not when the dialog is cancelled or closed, so a stale error from a prior failed attempt will resurface the next time the dialog opens with no new action taken. It also ships with no test asserting the new error paths actually work, relying only on "existing tests still pass." Recommend clearing errorMessage on cancel/close (e.g., in the Cancel handler or an effect keyed on `open`) and adding at least one test covering the success:false and catch-block message extraction before merge.

ivantha:
### Automated review summary

**Findings**
- **[nit/style]** `tensormap-frontend/src/containers/ProjectsPage/CreateProjectDialog.jsx:71` The inline onOpenChange callback's parameter is named `open`, shadowing the outer destructured `open` prop (used two lines above at `open={open}`). It works correctly due to lexical scoping, but two different `open` bindings with different meanings in the same closure hurts readability; renaming to e.g. `nextOpen` would be clearer.
- **[minor/quality]** `tensormap-frontend/src/containers/ProjectsPage/CreateProjectDialog.jsx:117` The 'clear error, then close' sequence is implemented twice: once in the Dialog's onOpenChange wrapper (lines 71-74, which handles Escape/overlay-click/close-icon) and again inline in the Cancel button's onClick (lines 117-120), which calls the raw onOpenChange prop directly instead of going through the wrapper. Both paths happen to stay in sync today, but a future edit to one is easy to forget in the other; extracting a single `closeDialog` handler used by both would remove the duplication.
- **[minor/test]** `tensormap-frontend/src/containers/ProjectsPage/CreateProjectDialog.jsx:57` The new test suite covers the success:false, network-error, detail-array, and no-message-fallback branches, but no test exercises the branch where `err.response.data.detail` is a plain string rather than an array (the `typeof err.response?.data?.detail === "string"` ternary arm), leaving that path unverified.

**Conventions**
- _Hygiene is otherwise excellent (root-cause writeup, itemized changes, verification notes) but the PR doesn't reference or link a tracked issue, though none appears to be filed for this bug._

PR #377 adds visible error feedback to CreateProjectDialog: an errorMessage state populated from either a non-success API response or a caught exception (extracting a message from the API's error envelope, FastAPI validation `detail`, or a generic fallback), rendered in a role="alert" div, and cleared on retry, cancel, or dialog close. The fix is well-grounded — the backend genuinely returns non-2xx JSON bodies matching the extraction logic — and ships with thorough Vitest coverage (8 new test cases, all passing in CI). Two minor polish items are worth a look: a shadowed `open` parameter in the onOpenChange callback, and duplicated close/clear-error logic between the Dialog wrapper and the Cancel button. Overall this is a clean, mergeable bug fix with no blocking issues.

Issue #378, opened 2026-07-03
Fresh SQLite setup fails — several migrations use PostgreSQL-only syntax


ashishsoni-ai:
## Bug
Following the README's local dev setup (SQLite fallback), the backend fails to start on a fresh database. Running `alembic upgrade head` errors out in two separate migrations that use PostgreSQL-specific syntax with no SQLite fallback.

## Reproduction
1. Fresh clone, `uv sync`
2. `uv run alembic upgrade head`

## Error 1 — `a728b2d08d59_initial_schema.py`

sqlite3.OperationalError: near "EXTENSION": syntax error
[SQL: CREATE EXTENSION IF NOT EXISTS "uuid-ossp"]

Also uses `now()` as a column server_default (Postgres-only) and `gen_random_uuid()` / `ON CONFLICT DO NOTHING` in a seed insert — none of which exist in SQLite.

## Error 2 — `6a7b8c9d0e1f_add_unique_constraint_on_data_process_file_id.py`

NotImplementedError: No support for ALTER of constraints in SQLite dialect. Please refer to the batch mode feature...
`op.create_unique_constraint` / `op.drop_index` need `op.batch_alter_table` on SQLite.

## Impact
New contributors following the documented SQLite setup path cannot start the backend at all.

## Fix
I've branched both migrations on `bind.dialect.name` so they run correctly on both SQLite and PostgreSQL, tested locally end-to-end (migrations complete, server starts, frontend loads projects successfully). Opening a PR shortly.

PR #379, opened 2026-07-03
fix: make migrations SQLite-compatible

ashishsoni-ai:
## Summary
Fixes #379

**Update:** Fixed the UUID-encoding bug flagged in review — the SQLite seed 
insert wrote a 36-char hyphenated UUID string instead of the 32-char 
hex-without-dashes format that sa.UUID() actually binds on SQLite, which 
broke id-based lookups (GET/PATCH/DELETE by id) despite the row showing up 
fine in list views. While verifying, found the identical pattern in 
`f4a8c2e91b37_seed_sample_project_pipeline.py` (hardcoded hyphenated UUIDs) 
plus a `gen_random_uuid()` call in its downgrade path that would crash on 
SQLite — fixed both. Tested upgrade and downgrade paths end-to-end on a 
fresh SQLite DB, confirmed stored UUIDs now match the ORM's expected format.

Fresh SQLite setups fail during `alembic upgrade head` due to PostgreSQL-only syntax in two migrations. This PR branches both on `bind.dialect.name` so they work correctly on both SQLite and PostgreSQL.

## Changes
- `a728b2d08d59_initial_schema.py`: guard `CREATE EXTENSION`, `now()` server defaults, and the `gen_random_uuid()`/`ON CONFLICT` seed insert behind a Postgres check; use `CURRENT_TIMESTAMP` and `INSERT OR IGNORE` for SQLite
- `6a7b8c9d0e1f_add_unique_constraint_on_data_process_file_id.py`: wrap `drop_index`/`create_unique_constraint` in `batch_alter_table` for SQLite (and the downgrade path)

## Testing
- Fresh SQLite DB: `alembic upgrade head` now completes cleanly through `merge_heads_001`
- Backend starts successfully (`Application startup complete`)
- Frontend loads `/projects` and shows the seeded sample project
- No changes to PostgreSQL code path (still uses original statements)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `tensormap-backend/migrations/versions/a728b2d08d59_initial_schema.py:155` The SQLite branch of the seed insert binds the project id as `str(uuid.uuid4())` (a 36-char hyphenated string) via raw SQL, bypassing the ORM's UUID bind-processor entirely. `Project.id` is declared as `Column(postgresql.UUID(as_uuid=True), primary_key=True, ...)` in app/models/project.py; on SQLAlchemy 2.0, postgresql.UUID is generalized from the core `Uuid` type, which falls back to hyphen-less CHAR(32) hex encoding on backends without native UUID support (SQLite) for every value the ORM binds or compares — this is exactly the format every real ORM-driven insert/lookup uses, including `db.get(Project, project_id)` (used by GET/PATCH/DELETE /project/{project_id}). Because the seeded row's stored id was written in the 36-char hyphenated form instead of that hex encoding, it will never match what those lookups bind for comparison: the seeded 'Sample Project' shows up fine in GET /project (a plain SELECT with no id filter) but returns 404 on any id-based fetch, update, or delete on a fresh SQLite database.
- **[minor/test]** `tensormap-backend/migrations/versions/6a7b8c9d0e1f_add_unique_constraint_on_data_process_file_id.py:64` The PR's stated testing only exercises `alembic upgrade head`; the new SQLite-only `downgrade()` branch (batch_alter_table drop_constraint/create_index here, plus the equivalent dialect branching in initial_schema.py) is never exercised, so a break in the rollback path on SQLite would not surface until someone actually needs to downgrade.

**Conventions**
- **PR title:** Reads like the raw branch name ("Fix/sqlite compatible migrations") rather than a Conventional Commits summary — no type prefix/colon; the commit's own "fix: make initial_schema and unique-constraint migrations SQLite-compatible" would make a clearer title.
- _Commit message and PR description are clear, well-scoped, and link the issue; only the PR title needs tightening to match the Conventional Commits style used in the commit itself._

This PR branches the two PostgreSQL-only Alembic migrations on bind.dialect.name, and manual testing confirms alembic upgrade head now completes on a fresh SQLite database as described in the linked issue, with the PostgreSQL path left untouched. However, the SQLite seed-insert writes the sample project's id as a raw hyphenated UUID string that bypasses the ORM's type encoding, while Project.id (typed postgresql.UUID) falls back to hyphen-less hex on SQLite for every ORM-driven read/write — so the seeded project will list correctly but 404 on any id-based GET/PATCH/DELETE. Fix the seed value's encoding (e.g. insert through the ORM, or match its hex-without-dashes format) before merging, and consider exercising the new SQLite downgrade branches too, since the stated testing only covered the upgrade direction.

ashishsoni-ai:
@ivantha Fixed the UUID encoding issue — SQLite branch now uses uuid.uuid4().hex 
(and a shared _uid() helper for the fixed seed UUIDs) to match sa.UUID()'s 
storage format. Also found and fixed the same bug pattern in 
seed_sample_project_pipeline.py, plus a gen_random_uuid() crash in its 
downgrade path. Ran the full downgrade→upgrade cycle on a fresh SQLite DB 
to confirm both directions work. Ready for another look whenever you have time!

ivantha:
### Automated review summary

**Findings**
- **[nit/style]** `tensormap-backend/migrations/versions/f4a8c2e91b37_seed_sample_project_pipeline.py:51` This file gates on `is_sqlite` (else-branch means Postgres) while the other two migrations touched in this same PR gate on `is_postgres` (else-branch means SQLite). The inconsistent polarity makes it easy to misextend one file's branch logic if a third dialect is ever added later, since the two conventions are not interchangeable.
- **[minor/test]** `tensormap-backend/migrations/versions/a728b2d08d59_initial_schema.py:24` No test or CI job exercises the SQLite path this PR fixes. `.github/workflows/test.yml` only provisions a `postgres:15` service for backend tests, so a future migration that reintroduces Postgres-only syntax (the exact failure mode behind issue #378) would pass CI silently until someone manually runs `alembic upgrade head` against SQLite again.

**Conventions**
- **Description:** Content is thorough (explains what changed, why, and how it was tested), but the summary line says 'Fixes #379', which self-references this PR instead of the real bug report. The commit message correctly says 'Fixes #378' (Fresh SQLite setup fails), so the description's link should be corrected or merging will not auto-close the actual issue.
- _Solid, well-documented change overall; just fix the issue-link typo in the description before merge._

This PR correctly branches three Alembic migrations on bind.dialect.name so alembic upgrade head and downgrade work on both SQLite and PostgreSQL, replacing Postgres-only syntax (CREATE EXTENSION, now(), gen_random_uuid(), unguarded ALTER for constraints and indexes) with SQLite-safe equivalents (CURRENT_TIMESTAMP, batch_alter_table, client-generated hex-encoded UUIDs). The UUID-encoding approach is technically sound given how SQLAlchemy's Uuid/UUID type stores values on SQLite, and the description states both upgrade and downgrade paths were tested end-to-end on a fresh SQLite database with no changes to the Postgres path. Before merging, fix the PR description's issue link ("Fixes #379" self-references this PR; the commit message correctly says "Fixes #378") so the real bug report auto-closes, and consider adding a CI job or test that runs migrations against SQLite so this class of regression is caught automatically next time.

PR #384, opened 2026-07-14
feat: Model Export Service (SavedModel, TFLite, ONNX)(Week 7 Deliverables)

kartikeyg0104:
## Summary

This PR adds a model export service so users can download their trained models in production-ready formats directly from the Training page. After a training job completes, the trained model is persisted and can be exported as **SavedModel** (zip), **TFLite**, or **ONNX** — with a pre-flight compatibility check for ONNX and clear guidance when an architecture isn't convertible.

### Backend

**Export service** (`app/services/model_export.py`)
- `export_savedmodel(job_id, model_name)` — loads `model.keras`, saves as TensorFlow SavedModel, zips it. Cached: returns the existing zip on subsequent calls.
- `export_tflite(job_id, model_name)` — converts via `TFLiteConverter`, writes `.tflite`. Cached the same way.
- `export_onnx(job_id, model_name, graph_ir)` — converts via `tf2onnx` with a proper input signature (handles single- and multi-input models).
- `validate_onnx_compatible(model, graph_ir)` — pre-flight check that flags LSTM/GRU with `return_sequences=True` on variable-length inputs, which tf2onnx can't convert. Users get an actionable message ("set a fixed shape in your Input layer") instead of a cryptic converter error.
- `get_export_formats(job_id, model_name, graph_ir)` — reports availability, size, and expiry for each format, plus ONNX support status.
- `cleanup_exports(retention_days, session)` — deletes export directories older than the retention window (default 7 days, configurable via `EXPORT_RETENTION_DAYS`).
- `delete_job_exports(job_id)` / `delete_model_exports(model_id, session)` — targeted cleanup used when a model is deleted.
- TensorFlow and tf2onnx are lazily imported so non-training processes don't pay the import cost.

**Export router** (`app/routers/export.py`) — two endpoints under `/api/v1/model/export`:
- `GET /model/export/{job_id}/formats` — available formats with metadata (size, expiry, ONNX compatibility issues).
- `GET /model/export/{job_id}?format=savedmodel|tflite|onnx` — streams the file with the right content type and filename; generates the export lazily on first request. Returns 404 if the job/model/keras file is missing, 400 with structured `onnx_unsupported` details (issues + suggested fallback) when ONNX conversion isn't possible. Updates `last_export_download_at` on the job as a background task.

**Integration points**
- `MetricsCallback.on_train_end` now saves the trained model to `exports/{job_id}/model.keras`, making completed jobs exportable (save failures are logged, never fail the run).
- App lifespan starts a background task that runs `cleanup_exports` every 6 hours and cancels it cleanly on shutdown.
- `delete_model_service` cascades export-directory deletion for all of a model's training jobs (best-effort; model deletion proceeds even if disk cleanup fails).
- New `get_session()` context manager in `app/database.py` for sessions outside FastAPI request scope.
- New dependency: `tf2onnx>=1.16`.

### Frontend

- **`ExportPanel`** (`src/components/export/ExportPanel.tsx`) — shown on the Training page when a job completes. Fetches available formats, renders a download button per format with size and expiry info, disables ONNX with the reported issues when the architecture isn't supported, and refreshes format metadata after each download.
- **Types** (`src/types/export.ts`) — `ExportFormat`, `FormatInfo`, `ExportFormatsResponse`, `ExportError`.
- `useTrainingMetrics` exposes `isCompleted`; `Training.jsx` uses it (via a new `onComplete` callback) to mount the panel once training finishes.
- New URL constant `BACKEND_MODEL_EXPORT`.

## Testing

`tests/test_model_export.py` — 13 tests covering:
- SavedModel export creates a zip and is cached on the second call
- TFLite export creates the file
- ONNX pre-flight validation passes for simple models and fails for RNNs with variable-length inputs
- ONNX export raises `ONNXUnsupportedError` with actionable issues for unsupported architectures
- 404 when no `model.keras` exists for the job
- `get_export_formats` response structure
- Retention cleanup deletes only directories older than `retention_days`
- Per-job and per-model (cascade) export deletion

The two ONNX conversion tests skip automatically when tf2onnx or its dependencies are unavailable in the environment.

The cleanup/cascade tests create the parent `model_basic` row before inserting `training_job` rows, so they pass against PostgreSQL in CI (which enforces the foreign key) as well as local SQLite.

```
11 passed, 2 skipped
```

Ruff lint and format checks pass across the backend.

## Notes

- Exports are generated lazily on first download and cached under `exports/{job_id}/`, so repeated downloads are instant and no disk is spent on formats nobody requests.
- Retention default is 7 days, configurable with the `EXPORT_RETENTION_DAYS` environment variable.
- ONNX limitations are surfaced to the user with a suggested fallback (SavedModel/TFLite) rather than failing opaquely.

### DIFF
--- a/SECURITY_FIXES.md
+++ b/SECURITY_FIXES.md
@@ -1,0 +1,141 @@
+# Security and Quality Fixes for Export Service
+
+This document summarizes the critical fixes applied to the model export service based on the automated code review feedback.
+
+## Fixed Issues
+
+### 1. [BLOCKING/SECURITY] Path Traversal Vulnerability (Fixed ✅)
+
+**Issue:** Model names were validated only for non-emptiness but used directly in file paths, allowing attackers to write files outside the intended directory.
+
+**Fix:**
+- Added `_sanitize_filename()` function in `model_export.py` that:
+  - Removes path separators (`/`, `\`)
+  - Removes parent directory references (`..`)
+  - Only allows safe characters: alphanumeric, underscore, hyphen, dot
+  - Prevents hidden files (starts with dot)
+  - Limits length to 100 characters
+- Applied sanitization in all export functions: `export_savedmodel()`, `export_tflite()`, `export_onnx()`, and `get_export_formats()`
+- Added test `test_path_traversal_prevention()` to verify the fix
+
+**Impact:** Prevents arbitrary file write attacks through crafted model names.
+
+---
+
+### 2. [MAJOR/BUG] Blocking I/O in Async Endpoints (Fixed ✅)
+
+**Issue:** Export endpoints called synchronous TensorFlow/tf2onnx functions directly from async handlers, blocking the event loop and freezing all other requests.
+
+**Fix:**
+- Wrapped all blocking export calls with `run_in_threadpool()` in `export.py`:
+  - `await run_in_threadpool(export_savedmodel, job_id, model_name)`
+  - `await run_in_threadpool(export_tflite, job_id, model_name)`
+  - `await run_in_threadpool(export_onnx, job_id, model_name, model.graph_ir)`
+
+**Impact:** Prevents freezing of the entire application during model exports; maintains responsiveness for all users.
+
+---
+
+### 3. [MAJOR/QUALITY] Unnecessary Model Loading (Fixed ✅)
+
+**Issue:** `get_export_formats()` loaded the full TensorFlow model just to call `validate_onnx_compatible()`, which never actually used the model parameter.
+
+**Fix:**
+- Modified `get_export_formats()` to pass `None` as the model parameter to `validate_onnx_compatible()`
+- Removed the expensive `tf.keras.models.load_model()` call
+- `validate_onnx_compatible()` only inspects `graph_ir`, not the model
+
+**Impact:** Significantly faster format listing (called on every ExportPanel mount and after each download).
+
+---
+
+### 4. [MINOR/BUG] Background Task Session Issue (Fixed ✅)
+
+**Issue:** Background task `update_download_timestamp()` used the request-scoped `db` session, which would be closed before the task ran.
+
+**Fix:**
+- Changed background task to use `get_session()` context manager
+- Creates a new session specifically for the background task
+- Properly queries, updates, and commits within the new session context
+
+**Impact:** Ensures `last_export_download_at` is reliably updated.
+
+---
+
+### 5. [MINOR/BUG] Frontend Filename Fallback (Fixed ✅)
+
+**Issue:** Fallback filename for SavedModel was `{modelName}.savedmodel` instead of `{modelName}.savedmodel.zip`, and cross-origin requests couldn't read the `Content-Disposition` header.
+
+**Fixes:**
+- Updated frontend fallback logic in `ExportPanel.tsx` to use `.savedmodel.zip` for SavedModel format
+- Added `expose_headers=["Content-Disposition"]` to CORS middleware in `main.py`
+
+**Impact:** Users always get the correct filename with proper extension, even in cross-origin setups.
+
+---
+
+### 6. [MINOR/QUALITY] Error Handling in Cascade Cleanup (Fixed ✅)
+
+**Issue:** `delete_model_exports()` would abort on first failure, leaving remaining export directories undeleted despite being described as "best-effort."
+
+**Fix:**
+- Wrapped individual `delete_job_exports()` calls in try-except
+- Log failures but continue processing remaining jobs
+- Return count of successfully deleted directories
+
+**Impact:** True best-effort cleanup; one failed deletion doesn't prevent cleanup of others.
+
+---
+
+### 7. [NIT/STYLE] Duplicated Export Path (Fixed ✅)
+
+**Issue:** `metrics_callback.py` hardcoded `Path("./exports/{job_id}")` instead of importing `EXPORTS_BASE`.
+
+**Fix:**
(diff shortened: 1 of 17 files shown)


ivantha:
### Automated review summary

**Findings**
- **[blocking/security]** `tensormap-backend/app/services/model_export.py:35` model_name is validated only for non-emptiness (ModelSaveRequest/DLModelConfig use Field(min_length=1); the DB column is Field(max_length=100, unique=True) — no charset restriction anywhere) yet is interpolated directly into export file paths, e.g. `zip_path = export_dir / f"{model_name}.savedmodel.zip"` (same pattern in export_tflite at line 65 and export_onnx at line 156, and again in get_export_formats). Saving a model named something like `../../../../tmp/pwned` and then exporting it makes `shutil.make_archive` write the generated file outside `exports/{job_id}/` to an attacker-chosen filesystem path, which is then streamed back via FileResponse — a path-traversal / arbitrary-file-write reachable through the ordinary save-model → train → export flow with no special privileges.
- **[major/bug]** `tensormap-backend/app/routers/export.py:99` export_savedmodel/export_tflite/export_onnx are synchronous functions that do blocking TensorFlow/tf2onnx work (tf.keras.models.load_model, save_model, TFLiteConverter.convert(), tf2onnx.convert.from_keras() — seconds to minutes for real models), but download_export is `async def` and calls them directly with no run_in_threadpool/anyio.to_thread.run_sync. Since the app wraps FastAPI with Socket.IO on a single event loop, one export request freezes every other request and every other user's live training-progress events until the conversion finishes.
- **[major/quality]** `tensormap-backend/app/services/model_export.py:262` get_export_formats — hit on every ExportPanel mount and again after each download via the frontend's fetchFormats() — does `model = tf.keras.models.load_model(model_path)` purely to call `validate_onnx_compatible(model, graph_ir)` (line 91), but that function's `model` parameter is never read in its body; only `graph_ir` is inspected (its own docstring even lists a second, unimplemented check — 'Unsupported tf.keras ops (tf.while_loop etc.)' — that would actually need `model`). Every formats-listing call pays for a full, blocking model deserialization for zero behavioral benefit.
- **[minor/bug]** `tensormap-backend/app/routers/export.py:139` update_download_timestamp, registered via background_tasks.add_task, mutates and commits `job` through the `db: Session = Depends(get_db)` request-scoped session. FastAPI's documented execution order runs a dependency's post-yield teardown (here, get_db's `with Session(engine) as session:` exiting) before background tasks execute, so this task always runs against a session whose context manager has already exited. This PR adds `get_session()` in app/database.py in this very commit, explicitly for 'getting a database session outside of FastAPI requests,' but doesn't use it here — the one call site that needs it — risking a silently-failed or erroring last_export_download_at update (the file download itself still succeeds either way).
- **[minor/bug]** `tensormap-frontend/src/components/export/ExportPanel.tsx:97` The download filename fallback `${modelName}.${format}` produces `{modelName}.savedmodel` (no .zip) for the SavedModel format, which doesn't match the backend's actual `{model_name}.savedmodel.zip` naming (export.py line 100). This isn't a rare edge case: app/main.py's CORSMiddleware never sets expose_headers, so Content-Disposition (not on the CORS-safelisted header list) is invisible to this axios blob request whenever frontend and backend are cross-origin — the project's own dev setup (ports 3300/4300) — meaning the wrong-extension fallback name is what typically ships rather than the header-derived one.
- **[minor/quality]** `tensormap-backend/app/services/model_export.py:359` delete_job_exports re-raises after logging on a shutil.rmtree failure, and delete_model_exports's `for job in jobs: if delete_job_exports(job.id): ...` loop doesn't catch that — so one failed directory deletion aborts the loop and leaves every remaining job's export directory undeleted, despite the PR description calling this cascade cleanup 'best-effort.' Because cleanup_exports's DB-driven path only considers jobs still present in TrainingJob, a directory orphaned this way after the model/jobs are removed from the DB becomes unreachable by the periodic retention sweep too.
- **[minor/test]** `tensormap-backend/tests/test_model_export.py:1` All 13 new tests exercise only the service-layer functions in model_export.py; nothing tests app/routers/export.py itself — not the 404s for a missing job/model, not the ONNX onnx_unsupported 400 response shape, not the last_export_download_at background-task update (the exact path flagged above), and not the FileResponse filename/media-type mapping. That router holds most of this PR's new request/response-shaping logic and has zero direct coverage.
- **[nit/style]** `tensormap-backend/app/callbacks/metrics_callback.py:119` `export_dir = Path(f"./exports/{self.job_id}")` duplicates `model_export.EXPORTS_BASE = Path("./exports")` instead of importing it. If the base export path is ever changed (e.g. made configurable via an env var), this call site would silently fall out of sync and every subsequent export would start failing with 'model.keras not found'.

**Conventions**

This PR adds a model export service (SavedModel, TFLite, ONNX) with a lazily-generated, cached export pipeline, a retention-based cleanup job, and a frontend export panel, backed by thorough tests and an unusually detailed PR description. The most important issue is a path-traversal / arbitrary-file-write hole: model names are validated only for non-emptiness (no charset restriction) yet get interpolated directly into export file paths, so a crafted model name can make the export functions write outside the intended exports directory. Separately, the new export endpoints and the periodic cleanup task call TensorFlow/tf2onnx synchronously from inside async handlers, blocking the shared Socket.IO event loop for every concurrent user during a conversion, and the formats-listing endpoint reloads the full model from disk to feed a validator that never reads it. Recommend fixing the path-traversal issue before merge, then the blocking-call and background-task session-reuse issues; the export router itself also has no direct test coverage.

kartikeyg0104:
@ivantha Have done the required changes...

PR #386, opened 2026-07-29
feat(week9): Interpretability Backend - Confusion Matrix, Feature Importance, Async Polling

kartikeyg0104:
## Summary

Implement confusion matrix + classification report, permutation feature importance with 202/200 async polling, regression analysis, and task-type routing. All computations run via `asyncio.to_thread` to avoid blocking the event loop. Results cached in `training_job.analysis_cache`.

### Confusion Matrix + Classification Report
- Loads trained model from `exports/{job_id}/model.keras`
- Replicates exact train/test split from `model_run._prepare_training_data()` (shuffle `random_state=42`, split at `training_split / 100`)
- Computes `sklearn.metrics.confusion_matrix` + `classification_report(output_dict=True)`
- Returns: confusion matrix, per-class precision/recall/f1, overall accuracy, class names, sample count
- Cached in `training_job.analysis_cache["confusion_matrix"]` after first computation

### Task-Type Routing
- Routes via `model_basic.model_type` → `ProblemType` enum → string
- Classification → confusion matrix, classification report, feature importance
- Regression → residuals, MAE, MSE (returns 400 for confusion matrix requests)

### Permutation Feature Importance (202 Polling)
1. **First request** → sets "computing" sentinel in cache → fires `asyncio.create_task()` → returns **202** `{"status": "computing"}`
2. **Subsequent requests** while computing → returns **202**
3. **After completion** → cached result → returns **200** with data
- Guardrails: `MAX_SAMPLES = 1000`, `MAX_FEATURES = 20` (top by variance)
- Keras model wrapped in sklearn-compatible `_KerasEstimatorWrapper` for `permutation_importance()`

### Async Execution
- All blocking sklearn/TensorFlow computations run via `asyncio.to_thread()` — never blocks the FastAPI event loop

## Tests

**20/20 passing** (6 scaffolding + 14 interpretability)

| # | Test | Verifies |
|---|------|----------|
| 1 | `test_confusion_matrix_shape` | 3-class → 3×3 matrix |
| 2 | `test_confusion_matrix_cached_on_second_call` | DB cache hit on 2nd call |
| 3 | `test_classification_report_has_precision_recall` | precision/recall/f1 per class + averages |
| 4 | `test_regression_returns_residuals` | Residuals array, MAE, MSE present |
| 5 | `test_confusion_matrix_not_available_for_regression` | 400 for regression jobs |
| 6 | `test_get_confusion_matrix_endpoint_200` | Full endpoint → 200 |
| 7 | `test_get_confusion_matrix_endpoint_cached` | 2nd request: `cached: true` |
| 8 | `test_analysis_requires_completed_job` | Pending job → 400 |
| 9 | `test_feature_importance_returns_202_first` | First call → 202 |
| 10 | `test_feature_importance_returns_200_after_compute` | Cached data → 200 |
| 11 | `test_feature_importance_capped_at_1000_samples` | 2000 rows → uses 1000 |
| 12 | `test_feature_importance_capped_at_20_features` | 30 features → capped at 20 |
| 13 | `test_feature_importance_cached_on_second_call` | Cache hit on repeat |
| 14 | `test_feature_importance_names_match_dataset` | Names match columns |

### DIFF
--- a/tensormap-backend/app/routers/analysis.py
+++ b/tensormap-backend/app/routers/analysis.py
@@ -5,20 +5,28 @@ Provides endpoints for:
 - Feature importance analysis
 - Paginated prediction results
 
-Week 8 scaffolding: All routes return 501 Not Implemented.
-Full implementation in Week 9.
+Routes to different analysis types based on model_basic.model_type:
+- classification (1): confusion matrix, classification report, feature importance
+- regression (2): residual plot, error histogram, feature importance (MSE-based)
+- image_classification (3): confusion matrix, classification report, per-class grid
 """
 
 from fastapi import APIRouter, Depends, HTTPException, Query
+from fastapi.responses import JSONResponse
 from sqlmodel import Session
 
 from app.database import get_db
 from app.models.training_job import TrainingJob, TrainingStatus
+from app.services.interpretability import InterpretabilityService
+from app.shared.enums import ProblemType
 from app.shared.logging_config import get_logger
 
 logger = get_logger(__name__)
 
 router = APIRouter(prefix="/model/analysis", tags=["analysis"])
+
+# Singleton service instance — stateless, safe to share across requests.
+_service = InterpretabilityService()
 
 
 def _verify_job_completed(job_id: str, session: Session) -> TrainingJob:
@@ -47,6 +55,40 @@ def _verify_job_completed(job_id: str, session: Session) -> TrainingJob:
     return job
 
 
+def get_analysis_type(job_id: str, session: Session) -> str:
+    """Returns 'classification', 'regression', or 'image_classification'.
+
+    Resolves the problem type by following training_job → model_basic → model_type.
+
+    Args:
+        job_id: Training job ID
+        session: Database session
+
+    Returns:
+        String identifying the analysis type
+
+    Raises:
+        HTTPException: 404 if job or model not found
+    """
+    from app.models.ml import ModelBasic
+
+    job = session.get(TrainingJob, job_id)
+    if not job:
+        raise HTTPException(status_code=404, detail="Training job not found")
+
+    model = session.get(ModelBasic, job.model_id)
+    if not model:
+        raise HTTPException(status_code=404, detail="Model not found")
+
+    type_map = {
+        ProblemType.CLASSIFICATION: "classification",
+        ProblemType.REGRESSION: "regression",
+        ProblemType.IMAGE_CLASSIFICATION: "image_classification",
+    }
+
+    return type_map.get(model.model_type, "classification")
+
+
 @router.get("/{job_id}/confusion-matrix")
 async def get_confusion_matrix(
     job_id: str,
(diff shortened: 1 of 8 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `tensormap-backend/app/routers/analysis.py:129` get_confusion_matrix's 400 response for regression jobs tells callers to "Use /residual-plot instead" (line 129), but no /residual-plot route, or any other regression-analysis route, is registered anywhere in this router (verified across the whole PR branch, not just this file). InterpretabilityService.compute_regression_analysis_async and _compute_regression_analysis_sync are fully implemented in interpretability.py and covered by test_regression_returns_residuals, but that test calls the private _compute_regression_analysis_sync method directly; nothing under app/routers ever calls compute_regression_analysis_async. Regression models currently have no working interpretability endpoint at all, and the pointer in the error message 404s.
- **[major/bug]** `tensormap-backend/app/services/interpretability.py:395` compute_feature_importance_async's in-progress check only tests for status == "computing" (line 395). When the background task in _compute_and_cache_feature_importance fails, it writes {"status": "failed"} to the sentinel (line 420), which no longer matches that check, so the very next call falls through, re-marks the sentinel "computing", and fires a brand-new asyncio.create_task computation (line 402). Any deterministic failure cause (missing model export, bad test data, etc.) makes the endpoint silently retry the full permutation-importance computation on every 2-second poll from the frontend hook, forever, with the failure never surfaced to the caller as anything other than an endless 202.
- **[major/bug]** `tensormap-backend/app/services/interpretability.py:402` asyncio.create_task(self._compute_and_cache_feature_importance(job_id, session)) discards the returned Task instead of storing it. Per asyncio's own documentation, the event loop only keeps a weak reference to untracked tasks, so this task can be garbage-collected mid-run, especially since the endpoint returns its 202 response immediately afterward with nothing else referencing the task. It also hands the task the FastAPI request-scoped db session (Depends(get_db) in analysis.py), which get_db()'s "with Session(engine) as session" block closes as soon as the endpoint returns; the detached task goes on to read and write through that same session from a separate worker thread (via asyncio.to_thread at line 413) well after its owning request has already completed.
- **[major/bug]** `tensormap-backend/app/services/interpretability.py:475` get_feature_importance (analysis.py) never checks analysis_type the way get_confusion_matrix does, so it runs unmodified on a regression job. _KerasEstimatorWrapper.predict() does np.argmax(preds, axis=1), which only makes sense for multi-column softmax output; a regression model's single-column prediction array makes this always return index 0, so permutation_importance(..., scoring="accuracy") silently returns near-zero, meaningless "importances" instead of erroring out like confusion-matrix does, or computing the MSE-based importance promised in this file's own module docstring.
- **[minor/quality]** `tensormap-backend/app/services/interpretability.py:175` _load_test_data (line 109) and _load_test_data_with_features (line 175) duplicate close to 60 lines of identical CSV-loading, dropna, shuffle, and train/test-split logic; the only real difference is whether class_names or feature_names is returned. Any future fix to that shared preprocessing (for example the split-index math or the categorical-encoding branch) has to be applied in both copies or they will silently drift apart.
- **[minor/test]** `tensormap-backend/tests/test_interpretability.py:372` The 14 new tests thoroughly cover the happy paths for confusion-matrix and feature-importance, but the regression_job fixture is only ever exercised against /confusion-matrix (line 278 of this file); nothing calls /feature-importance with a regression-typed job, and nothing drives feature_importance_status to "failed" and checks what the next poll does. Both bugs flagged above would pass this suite untouched.

**Conventions**

This PR replaces the Week 8 confusion-matrix and feature-importance stubs with real sklearn-backed computation, and the 14 new tests it adds pass cleanly along with ruff lint and format checks on all three changed backend files. The feature-importance 202/200 polling path has a real reliability gap: a failed background computation is never distinguished from "not started," so every later poll silently re-triggers the same expensive permutation-importance run forever, and that background task is fired through an unreferenced asyncio.create_task() carrying a request-scoped DB session that may already be torn down by the time it executes. Regression models can also hit /feature-importance directly, unlike /confusion-matrix which correctly rejects them, and will get back meaningless near-zero importances instead of an error; meanwhile the /residual-plot endpoint that the confusion-matrix error message points regression users to was never actually registered, even though the regression-analysis service method and its test are already written. I'd fix the failure-state handling and wire up (or remove references to) the regression route before merging, and consider deduplicating the two near-identical data-loading helpers while touching this file.

PR #389, opened 2026-08-09
feat(week11): Hyperparameter Tuning Service — Grid & Random Search

kartikeyg0104:
## Summary

Implement hyperparameter tuning with grid search and random search. A `tuning_session` orchestrates multiple training-job trials with different hyperparameter combinations. Trials execute sequentially (thread-safe), grid search is capped at 50 combinations, each trial has a 600s timeout, progress is emitted via Socket.IO, and the best trial's hyperparams can be applied back to the model.

### Grid Search
- Generates Cartesian product of all discrete parameter lists
- Capped at `MAX_GRID_COMBINATIONS=50` — returns 400 if exceeded
- Example: `{"optimizer": ["adam", "sgd"], "batch_size": [16, 32]}` → 4 combinations

### Random Search
- Supports three spec types:
  - **Discrete list** → uniform random choice
  - **`log_uniform`** → `exp(uniform(log(min), log(max)))` for learning rates
  - **`uniform`** → uniform sampling between min and max
- Generates exactly `max_trials` combinations (default 20)

### Sequential Trial Execution
- Trials run one-at-a-time via `asyncio.to_thread()` → synchronous loop
- Each trial creates a `training_job` row (all metrics independently queryable)
- Model config is temporarily modified per trial, then restored after completion
- Avoids thread pool exhaustion — no parallel training

### Trial Timeout
- `threading.Timer(TRIAL_TIMEOUT_SECONDS=600)` sets the job's DB status to CANCELLED
- Existing `CancellationCheckCallback` detects this at the next epoch boundary and sets `model.stop_training = True`

### Early Stopping
- If `early_stop_threshold` is set and a trial meets/exceeds it (respecting direction), the session stops early
- `early_stopped=True` flag set on the tuning session

### Duration Estimation
- Uses the most recent completed job's wall-clock time for the same model
- Falls back to `DEFAULT_TRIAL_ESTIMATE_SECONDS=120` if no history exists
- Returned in the 202 response as `estimated_seconds`

### Apply Best
- `POST /apply-best` copies `optimizer`, `epochs`, `batch_size` from the best trial's hyperparams to `model_basic`
- `learning_rate` is stored in job hyperparams but not persisted to `model_basic` (no such column)

### Socket.IO Rooms
- `subscribe_tuning` → join room `tuning:{session_id}`, receive catch-up state (current progress + completed trials)
- `unsubscribe_tuning` → leave room
- Progress events: `tuning_progress` (per-trial), `tuning_complete` (session done), `tuning_catchup` (late subscriber replay)

## API Endpoints

| Method | Endpoint | Status | Description |
|--------|----------|--------|-------------|
| POST | `/api/v1/model/tuning/{model_name}` | 202 | Start tuning session |
| GET | `/api/v1/model/tuning/{tuning_id}` | 200 | Session status, trials, best params |
| DELETE | `/api/v1/model/tuning/{tuning_id}` | 204 | Cancel session + running trial jobs |
| POST | `/api/v1/model/tuning/{tuning_id}/apply-best` | 200 | Copy best hyperparams to model config |

## Tests

**29/29 passing** (26 unit + 3 integration)

| # | Test | Verifies |
|---|------|----------|
| 1 | `test_grid_combinations_correct` | 3×3×2 = 18 combos generated correctly |
| 2 | `test_grid_exceeds_max_returns_400` | >50 combos → 400 error |
| 3 | `test_grid_empty_discrete_returns_single_empty` | No discrete params → single empty dict |
| 4 | `test_grid_single_param` | Single param with 3 values → 3 combos |
| 5 | `test_random_search_respects_max_trials` | Exactly `max_trials` combinations generated |
| 6 | `test_log_uniform_sampling_in_range` | Sampled LR always between min and max |
| 7 | `test_uniform_sampling_in_range` | Uniform sampling stays within bounds |
| 8 | `test_discrete_choice_values` | Discrete choices only produce values from list |
| 9 | `test_tuning_session_created_on_post` | POST creates row in DB with correct fields |
| 10 | `test_estimate_seconds_in_response` | POST 202 response includes `estimated_seconds` |
| 11 | `test_grid_search_post_400_too_many` | Grid >50 combos → 400 from endpoint |
| 12 | `test_unknown_model_404` | Non-existent model → 404 |
| 13 | `test_unconfigured_model_400` | Model without training config → 400 |
| 14 | `test_get_tuning_session` | GET returns session with trial info |
| 15 | `test_get_tuning_session_404` | Non-existent session → 404 |
| 16 | `test_cancel_tuning_sets_status` | DELETE sets status=cancelled |
| 17 | `test_cancel_tuning_cancels_running_jobs` | DELETE also cancels running trial jobs |
| 18 | `test_cancel_404` | Non-existent session → 404 |
| 19 | `test_cancel_completed_is_noop` | Cancelling completed session is no-op |
| 20 | `test_apply_best_updates_model_config` | apply-best copies optimizer/epochs/batch_size to model |
| 21 | `test_apply_best_not_completed_400` | apply-best on non-completed session → 400 |
| 22 | `test_apply_best_404` | Non-existent session → 404 |
| 23 | `test_early_stop_terminates_session` | Threshold met → early_stopped=True |
| 24 | `test_trial_timeout_cancels_job` | Timeout → job status=CANCELLED |
| 25 | `test_estimate_duration_from_last_job` | Last job 60s, 10 trials → estimate 600s |
| 26 | `test_estimate_duration_no_history` | No history → default 120s per trial |
| 27 | `test_full_tuning_flow` | Start → create jobs → GET → apply-best (integration) |
| 28 | `test_grid_search_creates_correct_trials` | Grid 2×2 → 4 trials (integration) |
| 29 | `test_cancel_and_recheck` | Cancel running → GET shows cancelled (integration) |

### DIFF
--- a/tensormap-backend/app/main.py
+++ b/tensormap-backend/app/main.py
@@ -22,7 +22,18 @@ from app.exceptions import (
     validation_exception_handler,
 )
 from app.middleware import RequestIDMiddleware, RequestLoggingMiddleware
-from app.routers import analysis, data_process, data_upload, deep_learning, export, health, layers, project, training
+from app.routers import (
+    analysis,
+    data_process,
+    data_upload,
+    deep_learning,
+    export,
+    health,
+    layers,
+    project,
+    training,
+    tuning,
+)
 from app.shared.logging_config import get_logger
 from app.socketio_instance import sio
 
@@ -155,6 +166,7 @@ app.include_router(layers.router, prefix=settings.api_base)
 app.include_router(training.router, prefix=settings.api_base)
 app.include_router(export.router, prefix=settings.api_base)
 app.include_router(analysis.router, prefix=settings.api_base)
+app.include_router(tuning.router, prefix=settings.api_base)
 
 # Wrap FastAPI with SocketIO so socket.io requests are handled,
 # and everything else passes through to FastAPI.
--- a/tensormap-backend/app/models/__init__.py
+++ b/tensormap-backend/app/models/__init__.py
@@ -3,6 +3,8 @@ from app.models.ml import ModelBasic, ModelConfigs
 from app.models.project import Project
 from app.models.training_job import TrainingJob, TrainingStatus
 from app.models.training_metric import TrainingMetric
+from app.models.tuning_session import TuningSession, TuningStrategy
+from app.models.tuning_session import TuningStatus as TuningSessionStatus
 
 __all__ = [
     "DataFile",
@@ -14,4 +16,7 @@ __all__ = [
     "TrainingJob",
     "TrainingMetric",
     "TrainingStatus",
+    "TuningSession",
+    "TuningSessionStatus",
+    "TuningStrategy",
 ]
--- a/tensormap-backend/app/models/tuning_session.py
+++ b/tensormap-backend/app/models/tuning_session.py
@@ -1,0 +1,114 @@
+"""Hyperparameter tuning session model.
+
+A tuning session orchestrates multiple training-job trials with different
+hyperparameter combinations.  Each trial creates a ``TrainingJob`` row so
+metrics are independently queryable.  The session tracks overall progress,
+the best-performing trial, and supports early stopping.
+"""
+
+import uuid as uuid_pkg
+from datetime import UTC, datetime
+from enum import StrEnum
+
+from sqlalchemy import JSON, Column, DateTime, ForeignKey, String
+from sqlalchemy import Enum as SAEnum
+from sqlmodel import Field, SQLModel
+
+
+class TuningStrategy(StrEnum):
+    """Search strategy for hyperparameter exploration."""
+
+    GRID = "grid"
+    RANDOM = "random"
+
+
+class TuningStatus(StrEnum):
+    """Lifecycle states for a tuning session.
+
+    A session moves PENDING -> RUNNING -> COMPLETED on the happy path, or to
+    FAILED (unrecoverable error) / CANCELLED (user requested a stop).
+    """
+
+    PENDING = "pending"
+    RUNNING = "running"
+    COMPLETED = "completed"
+    FAILED = "failed"
+    CANCELLED = "cancelled"
+
+
+def _tuning_status_values(enum_cls) -> list[str]:
+    return [member.value for member in enum_cls]
+
+
+def _tuning_strategy_values(enum_cls) -> list[str]:
+    return [member.value for member in enum_cls]
+
+
+class TuningSession(SQLModel, table=True):
+    """A hyperparameter tuning session that orchestrates multiple training trials.
+
+    Each session defines a search space and strategy (grid or random), spawns
+    training jobs for each trial, tracks the best result, and optionally stops
+    early when a threshold is reached.
+    """
+
+    __tablename__ = "tuning_session"
+
+    id: str = Field(
+        default_factory=lambda: str(uuid_pkg.uuid4()),
+        primary_key=True,
+        max_length=36,
+    )
+    model_id: int = Field(
+        sa_column=Column(
+            "model_id",
+            ForeignKey("model_basic.id", ondelete="CASCADE"),
+            index=True,
+            nullable=False,
+        )
+    )
+    strategy: TuningStrategy = Field(
+        sa_column=Column(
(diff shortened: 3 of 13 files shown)


ivantha:
### Automated review summary

**GSoC scope:** overlaps reserved GSoC scope (issue #120, full)

**Findings**
- **[major/bug]** `tensormap-backend/app/services/tuning_service.py:267` run_tuning_loop has no top-level try/except. An error during trial-combination generation (sample_random_combination indexes spec["min"]/spec["max"] unguarded, so a random-search spec with type log_uniform/uniform but a missing min or max raises KeyError) or any error in the per-loop session bookkeeping outside the try/except that wraps only _run_single_trial propagates out of the asyncio.to_thread task created in launch_tuning_session, whose result/exception is never retrieved (task.add_done_callback only discards it from the tracking set). Because tuning_session.status is already committed to RUNNING before this point, and the 'mark session complete' block at the end of the function never runs, the session is left stuck in RUNNING forever with no user-visible failure. TuningStatus.FAILED is declared in the model specifically for 'unrecoverable error' but is never assigned anywhere in this PR.
- **[major/bug]** `tensormap-backend/app/services/tuning_service.py:170` _apply_hyperparams_to_model/_restore_model_config cache each model's pre-trial optimizer/epochs/batch_size in self._original_configs, a plain dict on the module-level tuning_service singleton keyed only by model_name. Nothing in start_tuning checks for an already-running tuning session on the same model before launching a new one, and each session's trials execute in their own background thread via asyncio.to_thread. If two sessions run concurrently against the same model, the second session's _apply_hyperparams_to_model overwrites the first session's saved 'original' values with the first session's in-flight trial config, so _restore_model_config can write the wrong optimizer/epochs/batch_size back into model_basic for either session.
- **[minor/quality]** `tensormap-backend/app/schemas/tuning.py:26` TuningStartRequest.max_trials only enforces gt=0, with no upper bound, unlike grid search which generate_grid_combinations caps at MAX_GRID_COMBINATIONS=50. Trials run strictly sequentially in a single background thread with up to TRIAL_TIMEOUT_SECONDS=600s each, so a client-supplied max_trials of e.g. 100000 ties up that thread for an effectively unbounded duration with no server-side guard analogous to the grid-search cap.
- **[minor/quality]** `tensormap-backend/app/services/tuning_service.py:299` run_tuning_loop seeds every random-search session with np.random.default_rng(seed=42), a hardcoded constant rather than a per-session seed. Every random-search run over the same search_space, across different sessions and different users, samples the identical sequence of 'random' hyperparameter combinations in the identical order, defeating the exploration benefit a user would expect from re-running random search.
- **[minor/quality]** `tensormap-backend/app/schemas/tuning.py:17` TuningStartRequest.model_name is declared required (min_length=1), but start_tuning never reads body.model_name — only the model_name path parameter is used to look up the model. A client can send a body.model_name that doesn't match the URL and it is silently accepted and ignored rather than validated or rejected.
- **[nit/quality]** `tensormap-backend/app/schemas/tuning.py:32` TuningSessionResponse, TuningStartResponse, and TuningProgressEvent are fully-defined Pydantic response schemas, but no endpoint in routers/tuning.py declares response_model= against them — every endpoint returns a hand-built dict via _envelope() instead. These classes (and the trailing model_rebuild() call) are currently dead code with no effect on runtime validation or the generated OpenAPI schema.

**Conventions**

This PR adds a complete, well-tested backend hyperparameter-tuning service (grid and random search) with a new tuning_session table/migration, four REST endpoints, Socket.IO progress rooms, and a frontend API/socket service layer; all 29 new tests genuinely pass and lint, format, and the frontend build are all clean when verified locally against the PR branch. The main gap is error handling in the background tuning loop: run_tuning_loop has no top-level exception handling, so a malformed search space or any error outside the narrow per-trial try/except leaves the session stuck in RUNNING forever, and the FAILED status defined on the model is never actually set anywhere in the new code. There is also a concurrency risk: the pre-trial "original config" used to restore a model after each trial is cached in a singleton dict keyed only by model name, so two tuning sessions started concurrently against the same model can restore the wrong optimizer/epochs/batch_size into model_basic. Recommend wrapping run_tuning_loop in a try/except that marks the session FAILED, scoping the restore state per session (or guarding against concurrent sessions on the same model), and tidying a few minor API loose ends (unbounded max_trials, the unused model_name body field, and the unreferenced response schema classes) before merge.

PR #392, opened 2026-08-17
fix: reject disconnected canvas nodes instead of crashing with KeyError

### LINKED ISSUE #199: [Bug] model_generation.py crashes with KeyError for disconnected nodes and misidentifies input nodes as outputs
## Describe the bug
`model_generation()` in `app/services/model_generation.py` has two bugs:

**1. KeyError for disconnected nodes**
```python
output_ids = [n["id"] for n in model_params["nodes"] if n["id"] not in source_to_targets]
outputs = [keras_tensors[oid] for oid in output_ids]  # KeyError if node was never visited
```
If a node has no outgoing edges but was never visited during BFS (e.g. disconnected from the graph), it appears in `output_ids` but has no entry in `keras_tensors`, causing a `KeyError` crash.

**2. Input nodes misidentified as outputs**
`custominput` nodes have no outgoing edges in an incomplete graph, so they satisfy `n["id"] not in source_to_targets` and get included in `outputs`. This causes Keras to build a degenerate model where inputs equal outputs.

## Steps to Reproduce
1. Build a canvas with a disconnected node (not connected to the main graph)
2. Submit for validation
3. Observe `KeyError` crash in `model_generation()`

## Expected behavior
- Skip nodes not present in `keras_tensors` when collecting outputs
- Exclude `custominput` nodes from output identification

## Suggested Fix
```python
output_ids = [
    n["id"] for n in model_params["nodes"]
    if n["id"] not in source_to_targets
    and n["type"] != "custominput"
    and n["id"] in keras_tensors
]
```

### LINKED ISSUE #244: [Bug] Disconnected canvas nodes cause KeyError crash and input nodes misidentified as outputs in model_generation.py
**Describe the bug**

Two critical bugs in model_generation.py silently produce incorrect 
Keras models with no error shown to the user.

Bug 1 — KeyError crash on disconnected nodes:
If a user leaves any unconnected layer on the canvas, the BFS 
traversal never visits it. But the code still tries to look it 
up in keras_tensors, causing a hard KeyError crash with no 
helpful error message to the user.

Bug 2 — Input nodes misidentified as outputs:
The output detection logic uses:
output_ids = [n["id"] for n in model_params["nodes"] if n["id"] not in source_to_targets]

This incorrectly includes custominput nodes as outputs when they 
have no outgoing edges, silently producing a malformed Keras model.

**To Reproduce**

1. Open TensorMap at localhost:3300
2. Create a new project
3. Add an Input node and a Dense node on the canvas
4. Do NOT connect them with an edge
5. Save the model and click Train
6. Backend crashes with KeyError — no error message shown to user

**Expected behavior**
A clear, descriptive error message should tell the user exactly 
which nodes are disconnected, instead of crashing silently with 
a KeyError.

**Screenshots**
Not applicable — bug occurs in backend logic during model generation.

**Environment:**
OS: macOS (latest)
Browser: Chrome (latest)
Node.js version: 18+
Python version: 3.12+

**Additional context**
Affected file: tensormap-backend/app/services/model_generation.py
Affected lines: 65-70

Proposed fix:

# After BFS loop — validate all nodes were visited
unvisited_non_input = {
    nid for nid in nodes_by_id
    if nid not in visited
    and nodes_by_id[nid]["type"] != "custominput"
}
if unvisited_non_input:
    raise ValueError(
        f"Disconnected nodes detected: {unvisited_non_input}. "
        f"Please connect all layers before generating the model."
    )

# Fix output detection to exclude input nodes
output_ids = [
    n["id"] for n in model_params["nodes"]
    if n["id"] not in source_to_targets
    and n["type"] != "custominput"
]

Directly related to GSoC 2026 model generation reliability goal.

ashishsoni-ai:
## Description

`model_generation()` crashed with an unhandled `KeyError` whenever the ReactFlow graph contained a node that the BFS never reached, and silently built a malformed model when an input node had no outgoing edges.

Output detection used:

```python
output_ids = [n["id"] for n in model_params["nodes"] if n["id"] not in source_to_targets]
outputs = [keras_tensors[oid] for oid in output_ids]
```

An orphan node has no outgoing edges, so it is not in `source_to_targets` and is collected as an output — but the BFS never built a tensor for it, so the `keras_tensors[oid]` lookup raises a bare `KeyError` and the user gets a 500 with no indication of which node is at fault. The same rule also swept up `custominput` nodes with no outgoing edges, which *are* in `keras_tensors`, so instead of crashing they silently produced a model whose input was also its output.

This PR validates the graph before handing it to Keras:

- **No input node** → `Model has no input layer.`
- **Nodes unreachable from any input** → error naming each offending node ID. This also covers cycles, whose nodes never satisfy the "all sources visited" guard and were previously dropped from the model without warning.
- **Input nodes excluded from output detection**, plus an explicit error when that leaves the graph with no output layer.
- **Dangling input nodes** → error naming them. These are seeded into `visited` before the BFS runs, so the reachability check cannot catch them; without this Keras fails later with an opaque `` `inputs` not connected to `outputs` ``.

`ValueError` is the right signal here: both call sites in `services/deep_learning.py` already wrap `model_generation()` in `except ValueError` and return a 400 with the message, so these reach the UI as actionable feedback instead of a 500.

### Note on guard ordering

The four guards are deliberately ordered. The dangling-input check runs *after* the no-output check, not before it. A lone input node on an otherwise empty canvas satisfies both conditions, and running the dangling check first would make the no-output branch unreachable dead code. In the current order both errors are reachable and both are covered by a test.

Fixes #244
Fixes #199

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

Added `TestGraphConnectivityValidation` to `tests/test_model_generation.py` — 11 new cases covering: the exact reproduction from the issue (unconnected Input + Dense), an orphan alongside a valid path, a regression guard that the failure is `ValueError` and never `KeyError`, multiple orphans all named in one message, cycles, a graph with no input node, an empty graph, a dangling input node not being treated as an output, an input-only graph, and two guards that well-formed single- and multi-input graphs still build.

Also tightened the existing `test_single_input_no_edges`, which previously accepted either an exception or success to paper over a Keras 2/3 behaviour difference. That case is now deterministic and asserts the specific error.

- [x] Existing tests pass
- [x] New tests added
- [ ] Manual testing

**Backend suite:** ~469 passed, 5 skipped. `tests/test_model_generation.py` goes from 27 to 38 tests, all passing. Ruff `check` and `format --check` clean across the backend.

Two caveats, both confirmed pre-existing by stashing this branch's changes and re-running against an unmodified checkout:

1. `test_cp2_integration.py::test_training_jobs_list_endpoint` fails. It asserts a `started_at DESC` ordering over three rows inserted in the same instant, so the tie-break is nondeterministic. Fails identically on unmodified `main`, and is unrelated to model generation.
2. Running the whole suite in a single process hits `Fatal Python error: Aborted` during a lazy TensorFlow import on Windows (TF 2.16 / Python 3.12). This aborts at the identical line on an unmodified checkout, so it is an environment issue rather than a code one. Results above were collected by running each test file in its own process.

## Checklist

- [x] Tests pass locally

### DIFF
--- a/tensormap-backend/app/services/model_generation.py
+++ b/tensormap-backend/app/services/model_generation.py
@@ -36,18 +36,21 @@ def model_generation(model_params: dict) -> dict:
 
     nodes_by_id = {node["id"]: node for node in model_params["nodes"]}
 
+    input_nodes = [node for node in model_params["nodes"] if node["type"] == "custominput"]
+    if not input_nodes:
+        raise ValueError("Model has no input layer. Add an Input node to the canvas before training.")
+
     # BFS from input nodes to build Keras layers in topological order
     keras_tensors = {}
     visited = set()
     queue = []
 
-    for node in model_params["nodes"]:
-        if node["type"] == "custominput":
-            dims = [int(node["data"]["params"].get(f"dim-{i + 1}", 0) or 0) for i in range(3)]
-            dims = [d for d in dims if d != 0]
-            keras_tensors[node["id"]] = tf.keras.Input(shape=dims, name=node["id"])
-            visited.add(node["id"])
-            queue.append(node["id"])
+    for node in input_nodes:
+        dims = [int(node["data"]["params"].get(f"dim-{i + 1}", 0) or 0) for i in range(3)]
+        dims = [d for d in dims if d != 0]
+        keras_tensors[node["id"]] = tf.keras.Input(shape=dims, name=node["id"])
+        visited.add(node["id"])
+        queue.append(node["id"])
 
     while queue:
         current_id = queue.pop(0)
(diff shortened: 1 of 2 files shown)


ivantha:
### Automated review summary

**Findings**
- None flagged.

**Conventions**

This PR fixes a real bug: model_generation() crashed with a bare KeyError whenever the canvas graph had a node the BFS never reached (orphans, cycles), and silently built a malformed model when an input node had no outgoing edges. The fix adds four deliberately-ordered validation checks (no input, unreachable nodes, no output, dangling input) that raise ValueError with the offending node IDs named, which the existing call sites already convert into actionable 400 responses instead of a 500. Tracing the validation logic against the unmodified BFS/adjacency code turned up no correctness gaps, and the PR backs it with 11 new focused test cases plus a fix to a previously-flaky test that accepted either success or failure. Scope is tight (2 files, +205/-17), commit history and description are clean, and this looks safe to merge as-is.

PR #393, opened 2026-08-27
fix: [Bug] Input layer shape accepts zero and negative dimensions

### LINKED ISSUE #263: [Bug] Input layer shape accepts zero and negative dimensions
**Difficulty:** Easy
**Area:** full stack

## Description
Two of the three reported gaps have landed: the Dropout `rate` ParamSpec now carries `min=0.0` and `max=1.0` (`registry.py:383-390`), and `canSaveModel()` is registry-driven, validating every declared param and blocking the save on any violation (`Helpers.ts:32-71`, `:151`). What survives is the Input layer. Its `shape` ParamSpec is `required=True` but declares no `min`, unlike every other dimension-bearing int param in the registry (`units`, `filters`, `kernel_size`, `strides`, `pool_size`, `input_dim`, `output_dim` all use `min=1`; only Concatenate's `axis` is deliberately unbounded so it can be `-1`). Because the frontend validator enforces only the bounds the registry declares, `0` and negative dimensions pass the save guard. The IR schema does reject them (`InputParams.shape` has `gt=0`), but `model_validate_service()` swallows the IR conversion failure and falls back to the legacy generator, which filters zero dims out, so an Input of `0` silently becomes a scalar input instead of an error.

## Steps to Reproduce
1. Drop an Input node on the canvas and set its shape to `0` (or a negative value).
2. Connect a Dense layer downstream and save the model.
3. The save succeeds with no validation error; the backend logs an IR conversion warning and quietly falls back to the legacy generator.

## Expected vs Actual
- **Expected:** The Input shape field rejects anything below 1 with the standard `shape must be >= 1` message, and the save is blocked against that node.
- **Actual:** Zero and negative dimensions are accepted, saved, and silently reinterpreted downstream.

## Acceptance Criteria
- [ ] The `input` layer's `shape` ParamSpec in `registry.py` declares `min=1`, matching the other dimension params.
- [ ] `serialize_registry()` ships that `min` on `GET /layers`, so the frontend picks it up with no client-side special case.
- [ ] `canSaveModel()` rejects an Input node with shape `0` or a negative value and attributes the error to that node.
- [ ] Regression test in `tensormap-backend/tests/test_registry.py` asserting the `input` layer's `shape` spec has `min == 1`.
- [ ] Regression test in `tensormap-frontend/src/components/DragAndDropCanvas/tests/Helpers.test.ts` asserting a zero or negative Input shape fails `canSaveModel()`.

## Resources
- `tensormap-backend/app/layers/registry.py:114-121` - the `input` layer `shape` ParamSpec, the one dimension param with no `min`
- `tensormap-backend/app/layers/registry.py:130-136` - `dense.units`, the pattern to copy
- `tensormap-backend/app/layers/registry.py:383-390` - dropout `rate` clamping, already landed
- `tensormap-backend/app/layers/registry.py:484-494` - `serialize_registry()`, which forwards `min`/`max` to the frontend
- `tensormap-frontend/src/components/DragAndDropCanvas/Helpers.ts:44-55` - the int branch of `validateParamValue()`, which checks `min` only when the spec declares one
- `tensormap-backend/app/ir/schema.py:70` - `InputParams.shape` already has `gt=0`, so the registry is the layer that is out of step
- `tensormap-backend/app/services/deep_learning.py:141-142` - the IR conversion failure is logged and swallowed, so that `gt=0` never reaches the user
- `tensormap-backend/app/services/model_generation.py:49-51` - the legacy fallback drops zero dims, which is why a `0` shape fails silently

<details>
<summary>Original report by @Rakshi2609 (2026-03-26)</summary>

While the Layer Configuration Panel is functional, there are several validation gaps that could lead to invalid models being sent to the backend:

Dropout Rate: The rate input allows values outside the [0, 1] range (e.g., a user can type 5).
Input Layer: Model dimensions can currently be set to 0 or negative values, which are invalid for neural networks.
Save Validation: The canSaveModel utility does not explicitly check if the Dropout rate has been filled before allowing a save.
Proposed Fix:

Implement range clamping for the Dropout rate in the UI.
Enforce positive integers for Input layer dimensions.
Update Helpers.jsx to include Dropout validation in the save guards.

</details>

Mr-Neutr0n:
Fixes #263

Added `min=1` to the Input layer's `shape` ParamSpec in `registry.py` so `serialize_registry()` exposes the bound to the frontend and `canSaveModel()` rejects zero/negative shapes; added backend and frontend regression tests. (Issue text contained social/instructional content unrelated to the code fix — ignored per task rules.)

Could not run the suite locally: !!!!!!!!!!!!!!!!!!!!!!!!!!!. Fork CI needs a maintainer approval to run, so this branch has no test signal yet.

---
This change was prepared with AI assistance under human direction and review.


ivantha:
### Automated review summary

**Findings**
- **[nit/quality]** `tensormap-backend/app/services/model_generation.py:49` Fix is client-side only: it relies entirely on the registry min bound flowing through serialize_registry() -> canSaveModel()/validateParamValue() to block bad Input shapes. There's no corresponding backend-level guard, so a request that bypasses the UI (direct API call, stale cached frontend) can still hit the legacy generator's silent zero-dim filtering (dims are dropped with `d != 0`, negative dims aren't filtered at all) after the IR-path failure at deep_learning.py is only logged and swallowed rather than surfaced.

**Conventions**
- **Description:** Contains leftover automation artifacts that a maintainer can't parse — a garbled 'could not run the suite locally' line trailing off into a string of exclamation marks, and a cryptic aside about ignoring 'social/instructional content... per task rules' — even though the substantive part (what changed, why, and Fixes #263) is otherwise clear.
- **Commits (1):** The single commit message 'Fix #263: [Bug] Input layer shape accepts zero and negative dimensions' doesn't follow Conventional Commits — it uses a capitalized 'Fix #263:' issue-number prefix instead of a lowercase type prefix like 'fix:' (the PR title itself gets this right).
- _Tighten the commit message to Conventional Commits format and strip the leftover automation artifacts from the PR description before merge; the code change itself is minimal, correctly scoped, and well tested._

This is a minimal, correctly targeted fix: adding `min=1` to the Input layer's `shape` ParamSpec (tensormap-backend/app/layers/registry.py:118) is sufficient on its own, since the existing generic `serialize_registry()` and frontend `validateParamValue()`/`canSaveModel()` machinery already forwards and enforces any declared `min` — no other production code needed to change. New tests were added at both layers (a backend registry assertion, a parametrized frontend test for 0 and -5) and verifiably exercise the new bound. The only gap is that this closes the hole client-side only; a request that bypasses the UI can still reach the legacy generator's silent zero-dim handling, which is a pre-existing condition this PR doesn't regress but also doesn't close. Convention-wise, tidy up the commit message to a proper `fix:` prefix and clean the garbled automation artifacts out of the PR description before merging.

PR #395, opened 2026-08-27
fix(data): handle CSV metadata refresh failures

### LINKED ISSUE #119: [Bug] Preprocessing leaves DataFile.columns stale in the file list

Mr-Neutr0n:
Fixes #119

Added `refresh_data_file_columns_cache` to re-read and persist CSV headers after upload and preprocessing, fixing stale column lists in `GET /data/upload/file` when transformations change the header. CSV cache-read failures now still commit the `DataFile` record, while database commit failures roll back and propagate.

Validation:
- 52 upload, preprocessing, and cache tests passed.
- 384 non-slow, non-integration backend tests passed.
- Full local coverage remains blocked by TensorFlow aborting during real-model tests.

---
This change was prepared with AI assistance under human direction and review.


ivantha:
### Automated review summary

**Findings**
- **[blocking/bug]** `tensormap-backend/app/services/data_upload.py:138` In add_file_service's CSV branch the new DataFile row is only db.flush()'d (not committed) before refresh_data_file_columns_cache runs. That helper's except Exception swallows any failure from its own pd.read_csv call before it ever reaches db.commit() -- and since this call is the only commit point in that branch, a malformed/unreadable CSV now makes the endpoint return 201 'File saved successfully' while the flushed INSERT is rolled back when the request-scoped session closes (get_db in database.py neither commits nor rolls back explicitly, so Session.close() discards the pending transaction). The uploaded file is left on disk with no corresponding DataFile row at all.
- **[major/bug]** `tensormap-backend/app/services/data_upload.py:29` refresh_data_file_columns_cache uses a bare except Exception, which now also covers failures from its own db.commit() call (e.g. IntegrityError/OperationalError), not just the narrower CSV-read failures the old code targeted (pd.errors.ParserError, OSError, UnicodeDecodeError, MemoryError). On a commit failure it just logs a generic warning and returns without calling db.rollback(), leaving the session's transaction aborted for any later use in the same request. Because file.columns/file.row_count are mutated on the in-memory object before that failing commit, the preprocess_data call site can also report 200 success while the exact stale-cache bug this PR is fixing reappears whenever the commit itself fails.

**Conventions**
- **PR title:** Title is truncated mid-word ('...stale in the fil') and stacks a redundant '[Bug]' tag on top of the Conventional Commits 'fix:' prefix.
- **Commits (1):** Commit message 'Fix #119: [Bug] Preprocessing leaves DataFile.columns stale in the fil' doesn't follow Conventional Commits (capitalized 'Fix', issue-number prefix instead of a lowercase 'fix: ...' summary) and is truncated mid-word, same as the PR title.
- _Both the PR title and the single commit message look cut off mid-sentence and don't follow the repo's lowercase Conventional Commits style; worth renaming to a complete 'fix(scope): ...' summary before merge._

PR #395 adds a shared refresh_data_file_columns_cache helper so DataFile.columns/row_count are re-read from disk after both upload and preprocessing, which does fix the reported staleness on the happy path. However, in add_file_service's CSV branch the new row is only flushed (not committed) before that helper runs, so a CSV read failure inside the helper is swallowed silently and the row is never committed -- the endpoint still reports 201 success while the upload is effectively lost. The helper's blanket except Exception also wraps its own db.commit() call with no db.rollback(), which can leave the session in an aborted state and lets the very staleness bug this PR targets reappear whenever that commit fails. Separately, the PR title and its one commit message are both truncated and don't follow the repo's Conventional Commits style.

Mr-Neutr0n:
Thanks for the detailed review. I pushed `1240ef7` and updated the title to `fix(data): handle CSV metadata refresh failures`.

The CSV read is now separate from the database write. If pandas cannot read the file, `add_file_service` commits the `DataFile` row without cached metadata. The list path can backfill it later. If persisting the metadata fails, the session rolls back and the error propagates instead of returning success.

I added regression coverage for the cache-read failure and the rollback path.

Validation:
- 52 focused upload, preprocessing, and cache tests pass.
- 384 non-slow, non-integration backend tests pass.
- Ruff check and format pass.

The full suite still aborts in this local environment when a real TensorFlow model test loads. That is outside the changed upload path. No CI checks are reported for the branch yet.

The branch is ready for re-review.

Mr-Neutr0n:
Closing as superseded by #397, which was merged for the same issue and closed #119. Its implementation reads CSV metadata before creating the upload record, so the upload transaction concern raised here does not carry over. Thanks for the review and to @afonsojanu for landing the fix.

PR #396, opened 2026-08-27
fix: raise a descriptive error instead of KeyError in _map_template

### LINKED ISSUE #195: [Bug] _map_template raises KeyError on an unknown model_type

afonsojanu:
## Description

`_map_template()` indexed its `options` dict directly (`options[problem_type_id]`), so any `model_type` outside `ProblemType` escaped as a bare `KeyError` — including `None`, which `ModelBasic.model_type` (`int | None`) explicitly allows. `get_code_service()` only catches `ValueError` and `CodeGenerationError`, so a request for a model with no problem type set died as an opaque 500 instead of a 4xx naming the unsupported type.

Fixes #195

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [x] Existing tests pass
- [x] New tests added
- [ ] Manual testing

Added `test_none_model_type_raises_value_error_naming_it` and `test_unrecognised_model_type_raises_value_error_naming_it` to `TestMapTemplate`, per the issue's acceptance criteria. Confirmed both fail with a raw `KeyError` against the old code (reverted the fix locally, re-ran) and pass against the fix.

```
uv run pytest tests/test_null_checks.py tests/test_code_generation_service.py -v
# 16 passed
uv run ruff check app/services/code_generation.py tests/test_code_generation_service.py
# All checks passed!
```

The three existing `TestMapTemplate` cases for the valid `ProblemType` values still pass unchanged.

`get_code_service()` needed no change: it already catches `ValueError` and returns a 400, so raising `ValueError` from `_map_template()` is enough to turn the 500 into a 4xx.

**Note on the full suite:** `uv run pytest tests/` crashes with a native `Fatal Python error: Aborted` inside real TensorFlow's C extension import, in `test_migration.py` and `test_interpretability.py` — unrelated to this change. Confirmed by reverting my diff and re-running: the identical crash happens on unmodified `main` in this sandbox (Python 3.12, macOS arm64), so it's a pre-existing environment issue, not something this PR introduces. All tests in the two files this PR touches pass cleanly, both alone and together.

N/A — backend-only change, no UI.

## Checklist

- [ ] I have added/updated documentation as needed (no user-facing docs cover this internal error path)
- [x] Tests pass locally


ivantha:
### Automated review summary

**Findings**
- None flagged.

**Conventions**

Clean, well-scoped fix: _map_template() now catches the KeyError from indexing the options dict (including the None case, since ModelBasic.model_type is int | None) and re-raises a ValueError naming the offending value and listing the valid IDs; get_code_service() already turns ValueError into a 400 via its existing except clause, so no caller-side change was needed. Two new regression tests cover None and an out-of-range int (99), and the three existing valid-mapping tests are untouched. Only one call site exists (generate_code()), so there's no risk of a missed caller, and ruff passes clean on both touched files. One non-blocking style note: the "Valid: 1, 2, 3" message surfaces raw enum ints rather than names, so a reader without ProblemType memorized would need to cross-reference enums.py to know what 1/2/3 mean.

ivantha:
@afonsojanu Please fix the lint errors.

PR #397, opened 2026-08-27
fix: refresh DataFile.columns cache after preprocessing

### LINKED ISSUE #119: [Bug] Preprocessing leaves DataFile.columns stale in the file list

afonsojanu:
## Description

`preprocess_data()` overwrites the CSV on disk when a transformation changes its header (Drop Column, One Hot Encoding, Categorical to Numerical), but never refreshed the `DataFile.columns` cache that `get_all_files_service` prefers whenever it is not `None`. `GET /data/upload/file` kept serving the pre-transform column list until the file was deleted and re-uploaded.

Fixes #119

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## How Has This Been Tested?

- [x] Existing tests pass
- [x] New tests added
- [ ] Manual testing

Extracted the shared header-reading logic (previously duplicated between `add_file_service`'s upload-time caching and `get_all_files_service`'s lazy backfill) into `app/shared/csv_columns.read_csv_columns()`, and call it from `preprocess_data()` after the atomic `os.replace()` succeeds.

Added `test_drop_column_refreshes_the_columns_cache`, which drops a column via `preprocess_data`, asserts the `DataFile.columns` attribute is refreshed, then calls `get_all_files_service` and asserts the file-list response reflects the new columns too — per the issue's acceptance criteria. Confirmed it fails against the old code (reverted the fix locally via `git stash`, re-ran): the stale `"species"` column was still present in both places.

```
uv run pytest tests/test_data_process_service.py tests/test_data_upload_service.py tests/test_data_process.py tests/test_data_process_backup.py tests/test_data_process_params.py tests/test_data_process_correlation.py -v
# 68 passed
uv run ruff check app/services/data_process.py app/services/data_upload.py app/shared/csv_columns.py tests/test_data_process_service.py
# All checks passed!
```

The lazy backfill branch in `get_all_files_service` (data_upload.py:160-174) is unchanged — it already recomputes on demand and was not the source of staleness, so I left it alone per the issue's explicit acceptance criterion that it "still works for rows uploaded before the cache existed."

N/A — backend-only change, no UI.

## Checklist

- [ ] I have added/updated documentation as needed (no user-facing docs cover this internal cache)
- [x] Tests pass locally


ivantha:
### Automated review summary

**Findings**
- **[minor/bug]** `tensormap-backend/app/services/data_process.py:441` The cache-refresh db.commit() at the end of preprocess_data is not wrapped in its own try/except, so if it raises (e.g. a transient DB error) it falls through to the function's generic except Exception handler and returns a 500 'Error preprocessing data' even though the CSV on disk was already successfully rewritten (with a .bak backup made) moments earlier by the same call. A client that retries after such a false failure could 422 on an already-applied transform, or silently double-apply a non-idempotent one like Log Transform.

**Conventions**

The fix correctly closes the gap: after preprocess_data overwrites the CSV, it now re-reads the header via a new shared read_csv_columns() helper (also reused by upload-time caching, removing prior duplication) and persists it onto DataFile.columns, so get_all_files_service no longer serves the pre-transform column list. The new test exercises the full path end-to-end — drop a column via preprocess_data, then assert get_all_files_service reflects it — and the PR description documents that it fails against the pre-fix code. One narrow edge case: the final db.commit() for the cache update isn't isolated from the function's broad except Exception, so a DB error at that exact point reports the whole preprocess as failed despite the on-disk transformation having already succeeded. Otherwise the change is small, well-scoped, and safe to merge.

PR #401, opened 2026-09-02
fix: harden model save validation against malformed graphs (KeyError, missing IR validation)

### LINKED ISSUE #400: Bug: model save endpoints crash with HTTP 500 on malformed graphs (edge to nonexistent node, missing IR validation)
## Summary

The model save/validate endpoints (`POST /api/v1/model/save`, `POST /api/v1/model/validate`) can produce HTTP 500 Internal Server Errors or opaque Keras error messages for several classes of malformed graph that should produce a clean HTTP 400 with a descriptive message.

### 1. Edge to nonexistent node causes KeyError → HTTP 500

An edge that references a target node not present in the graph's `nodes` list causes a bare `KeyError` in both the legacy generator (`model_generation.py`) and the registry-driven generator (`tensorflow_generator.py`). The service layer catches only `ValueError`, so the `KeyError` propagates as a 500.

**Reproduction:** Send a payload with `edges: [{"source": "n1", "target": "ghost"}]` where `"ghost"` is not in `nodes`.

### 2. `validate_ir_graph()` is never called in the model save/validate flow

The `validate_ir_graph()` function in `app/ir/schema.py` detects cycles, edges to non-existent nodes, invalid Concatenate arity, and missing input nodes — but it is only called by the separate `/layers/validate-graph` endpoint, never by the main model save/validate endpoints. Structural graph errors are only caught later (if at all) when Keras attempts to build the model, producing opaque error messages.

### 3. `model_validate_service`/`model_save_service` catch only `ValueError`

The legacy generator fallback catches only `ValueError`, but the generator can also raise `KeyError` (issue 1) or `TypeError` (from wrong parameter types). These propagate as 500 errors.

## Impact

Malformed graph requests can crash the endpoint with a 500 error, making it harder for users to diagnose and fix broken models.

## Proposed fix

- Add existence checks in both generators before looking up `nodes_by_id[target_id]` and raise descriptive `ValueError`/`TensorFlowGeneratorError` instead.
- Widen the caught exception types in both service functions to include `KeyError` and `TypeError`.
- Call `validate_ir_graph()` after `reactflow_to_ir()` conversion and surface structural errors as 400 responses. Multi-input models (a supported feature) are excluded from this check.
- Add regression tests in `test_model_generation.py` for edge-to-nonexistent-node in both generators.

ashishsoni-ai:
## Summary

Fixes #400 — the model save/validate endpoints could return HTTP 500 or opaque Keras errors on malformed graphs instead of a clean HTTP 400.

### Backend

- **`model_generation.py`** — reject edges that reference nodes missing from the graph with a clear `ValueError` instead of crashing with a bare `KeyError`.
- **`tensorflow_generator.py`** — same guard in the registry-driven generator, raising `TensorFlowGeneratorError`.
- **`deep_learning.py`** — `model_validate_service` and `model_save_service`:
  - Widen the legacy-fallback exception handling to also catch `KeyError` and `TypeError` so malformed input returns a clean 400.
  - Call `validate_ir_graph()` after ReactFlow → IR conversion and surface structural errors (cycles, edges to unknown nodes, invalid Concatenate arity, missing input) as 400 responses instead of opaque Keras failures. Multi-input models remain supported (the `multiple input nodes` IR check is excluded via `_blocking_ir_errors`).

### Tests

- **`test_model_generation.py`** — regression tests for edge-to-nonexistent-node in both the legacy generator and the IR generator, asserting a descriptive error (not `KeyError`).

## Verification

- `test_model_generation.py`, `test_api_integration.py`, `test_cp1.py`, `test_graph_json.py` → 112 passed
- `ruff check` clean

## Related

- Closes #400

### DIFF
--- a/tensormap-backend/app/generators/tensorflow_generator.py
+++ b/tensormap-backend/app/generators/tensorflow_generator.py
@@ -113,6 +113,13 @@ class TensorFlowGenerator:
                 visited.add(target_id)
                 queue.append(target_id)
 
+                # An edge may reference a target node that does not exist in the
+                # graph. Reject it with a clear error instead of a bare KeyError.
+                if target_id not in nodes_by_id:
+                    raise TensorFlowGeneratorError(
+                        f"Edge {current_id} -> {target_id} references a node that does not exist in the graph"
+                    )
+
                 node = nodes_by_id[target_id]
 
                 # Gather input tensors
--- a/tensormap-backend/app/services/deep_learning.py
+++ b/tensormap-backend/app/services/deep_learning.py
@@ -75,6 +75,21 @@ def _extract_graph(payload: dict) -> dict | None:
     return payload
 
 
+def _blocking_ir_errors(graph_ir) -> list[str]:
+    """Return IR-graph validation messages that should block model saving.
+
+    Structural errors (cycles, edges to unknown nodes, invalid merge arity,
+    missing input) are surfaced early with clear messages instead of being
+    deferred to opaque Keras failures. Multi-input models are a supported
+    feature (both generators handle them), so the "multiple input nodes"
+    check is intentionally excluded.
+    """
+    from app.ir.schema import validate_ir_graph
+
+    errors = validate_ir_graph(graph_ir)
+    return [e.message for e in errors if "multiple input nodes" not in e.message]
+
+
 def _validate_graph_size(graph: dict | None) -> str | None:
     """Return an error message if the serialised graph exceeds the size limit, else None."""
     if graph is None:
@@ -138,6 +153,12 @@ def model_validate_service(db: Session, incoming: dict, project_id: uuid_pkg.UUI
         graph_ir = reactflow_to_ir(incoming["model"])
         graph_ir_data = graph_ir.model_dump()
         logger.debug("Successfully converted ReactFlow to IRGraph for model validation")
+
+        # Surface structural graph errors early with clear messages instead of
+        # opaque Keras failures later on.
+        blocking = _blocking_ir_errors(graph_ir)
+        if blocking:
+            return _resp(400, False, "; ".join(blocking))
     except Exception as e:
         logger.warning("Failed to convert ReactFlow to IRGraph (falling back to legacy generator): %s", str(e))
 
@@ -160,7 +181,7 @@ def model_validate_service(db: Session, incoming: dict, project_id: uuid_pkg.UUI
     if model_generated is None:
         try:
             model_generated = model_generation(model_params=incoming["model"])
-        except ValueError as e:
+        except (ValueError, KeyError, TypeError) as e:
             return _resp(400, False, str(e))
 
     try:
@@ -263,6 +284,12 @@ def model_save_service(db: Session, incoming: dict, model_name: str, project_id:
         graph_ir = reactflow_to_ir(incoming)
         graph_ir_data = graph_ir.model_dump()
         logger.debug("Successfully converted ReactFlow to IRGraph for model save")
+
+        # Surface structural graph errors early with clear messages instead of
+        # opaque Keras failures later on.
+        blocking = _blocking_ir_errors(graph_ir)
+        if blocking:
+            return _resp(400, False, "; ".join(blocking))
     except Exception as e:
         logger.warning("Failed to convert ReactFlow to IRGraph (falling back to legacy generator): %s", str(e))
 
@@ -285,7 +312,7 @@ def model_save_service(db: Session, incoming: dict, model_name: str, project_id:
     if model_generated is None:
         try:
             model_generated = model_generation(model_params=incoming)
-        except ValueError as e:
+        except (ValueError, KeyError, TypeError) as e:
             return _resp(400, False, str(e))
 
     try:
--- a/tensormap-backend/app/services/model_generation.py
+++ b/tensormap-backend/app/services/model_generation.py
@@ -65,6 +65,15 @@ def model_generation(model_params: dict) -> dict:
             visited.add(target_id)
             queue.append(target_id)
 
+            # An edge may reference a target node that does not exist on the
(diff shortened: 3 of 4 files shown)


ivantha:
### Automated review summary

**Findings**
- **[major/test]** `tensormap-backend/app/services/deep_learning.py:159` The new `_blocking_ir_errors` gating inside `model_validate_service`/`model_save_service` (computed at line 159, mirrored at line 290) — the service-level 400-on-structural-error behavior the PR description actually advertises — has no test exercising it. The tests added in `test_model_generation.py` only call the lower-level `model_generation()` and `TensorFlowGenerator.build_model()` directly; a wiring regression between `validate_ir_graph()` and the `model_validate_service`/`model_save_service` 400 response (e.g. the blocking check being skipped, miscalled, or the filter excluding more than intended) would pass this PR's test suite undetected.
- **[minor/quality]** `tensormap-backend/app/services/deep_learning.py:184` The fallback except clause was widened from `except ValueError` to `except (ValueError, KeyError, TypeError)` (also at line 315) with no logging call added alongside it. A genuine internal bug in `model_generation()` that happens to surface as a `KeyError`/`TypeError` unrelated to malformed user input is now silently converted to a 400 response with the raw exception text and no server-side trace of what actually failed, making such regressions harder to notice in production.
- **[nit/quality]** `tensormap-backend/app/services/deep_learning.py:90` `_blocking_ir_errors` excludes the 'multiple input nodes' validation result by substring-matching the free-text `IRValidationError.message` (`"multiple input nodes" not in e.message`) rather than a structured error code/type — `IRValidationError` only carries `message`/`node_id`/`field_name`, so this works today but silently stops excluding multi-input graphs if that message wording in `app/ir/schema.py` ever changes.

**Conventions**

This PR closes a real gap: both the legacy and IR-based generators could raise a bare `KeyError` when an edge references a node missing from the graph, and `model_validate_service`/`model_save_service` now call `validate_ir_graph` up front so structural IR errors (cycles, bad merge arity, missing input, dangling edges) surface as clean 400s instead of opaque Keras failures later. The new guards in `model_generation.py` and `tensorflow_generator.py` are correctly placed ahead of the dict lookups they protect, and the added tests exercise both generator paths directly with the right exception types. The weaker spot is that the service-layer wiring this PR actually advertises — `_blocking_ir_errors` inside `model_validate_service`/`model_save_service` — has no test of its own, so a regression there wouldn't be caught by this PR's suite; widening the fallback except clause to `(ValueError, KeyError, TypeError)` without adding any logging is also worth a second look, since it would now silently return a genuine internal bug as a 400 with no server-side trace.

ivantha:
### Automated review summary

**Findings**
- **[minor/quality]** `tensormap-backend/app/services/deep_learning.py:184` The fallback except clause was widened to catch KeyError and TypeError (same pattern repeated at line 316) and returns str(e) as the 400 message. This does stop the crash, but for KeyError cases other than the one new explicit check (dangling edge target in model_generation.py:71), the message is just the raw missing-key repr (e.g. a node missing data.params.activation at model_generation.py:165-192 would surface as the message "'activation'"), which isn't the 'clear error messages instead of opaque failures' the PR describes. It also means a genuine TypeError from a real bug elsewhere in generation would be silently reported as a 400 malformed-graph error, and logger.warning(str(e)) here doesn't include a traceback (no exc_info=True), making that case harder to tell apart from real user error during triage.
- **[nit/quality]** `tensormap-backend/app/services/model_generation.py:71` The new guard only checks that an edge's target_id exists in nodes_by_id. A dangling source_id (edge pointing from a deleted node) isn't given the same explicit, descriptive check in this legacy path; it instead falls through silently (the source is never in `visited`, so the target is never processed) and surfaces later as the generic 'not connected to any input layer' unreachable-nodes error rather than a message naming the actual bad edge. Minor inconsistency only — it doesn't crash, and the IR path (validate_ir_graph) already covers both source and target dangling references.

**Conventions**

This PR closes a real gap: a ReactFlow graph with an edge pointing at a deleted/nonexistent node previously crashed model save/validate with a bare KeyError or an opaque downstream Keras failure. Now both the legacy and IR-based generators raise descriptive errors, and the service layer runs the existing validate_ir_graph() structural checks up front and returns a clean 400 instead. The change is well tested (43 tests in the touched file, 113 passing across the four suites named in the description) and ruff-clean, verified locally. The one gap worth a follow-up: the fallback except clause was widened to KeyError/TypeError broadly, so malformed-graph cases outside the specific dangling-edge scenario (e.g., a node missing data.params.activation) still surface as a raw exception string rather than a clear message, and a genuine TypeError bug elsewhere could get silently reported as user error since the warning log has no traceback.

ivantha:
@ashishsoni-ai Please resolve the conflicts.

ivantha:
### Automated review summary

**Findings**
- **[minor/quality]** `tensormap-backend/app/services/deep_learning.py:121` The KeyError/TypeError-to-500 and ValueError-to-400 exception handling around the legacy model_generation() call is duplicated nearly verbatim between model_validate_service and model_save_service (and the new `_blocking_ir_errors` gating block is likewise duplicated). A shared helper would avoid the two copies drifting apart on the next change.

**Conventions**
- **Commits (4):** includes a merge commit and a follow-up "address automated review" fixup commit that should be squashed into the main fix commit before merge
- _Otherwise clean, well-tested, and each commit title is properly Conventional-Commits-formatted._

This PR closes a real gap where malformed graphs (edges pointing to nonexistent nodes, nodes missing required layer params) could crash with a bare KeyError/TypeError or surface an opaque 500 instead of a clean 400. The fix adds endpoint validation in both the legacy and IR-based generators, tightens exception handling in the two service functions to distinguish user-input errors (400) from genuine internal bugs (500, logged with traceback), and ships thorough regression tests covering both generators and both services. The only nit is that the validation-gating and exception-handling logic is duplicated between model_validate_service and model_save_service, which is worth extracting to a shared helper at some point but isn't blocking.

PR #403, opened 2026-09-02
fix: _estimate_param_count now handles scalar input shape (first Dense weights omitted)

### LINKED ISSUE #402: Bug: model architecture parameter count is wrong for scalar input shapes (first Dense weights omitted)
## Summary

`GET /api/v1/model/architecture/{id}?include_stats=true` reports an incorrect parameter count for models whose input shape is stored as a scalar integer (which is how the IR schema serializes it).

## Root cause

`_estimate_param_count` in `app/services/deep_learning.py` reads the input layer shape and only handles a `list`:

```python
shape = node_params.get("shape", [])
if isinstance(shape, list) and len(shape) > 0:
    prev_shape = shape[-1] if shape[-1] is not None else 128
```

But the IR schema (`InputParams.shape`) is an `int` (`shape: int = Field(..., gt=0, ...)`). When the shape is a scalar int (e.g. `10`), `isinstance(shape, list)` is `False`, so `prev_shape` stays `None` and **the first Dense layer's weights are never counted**.

**Reproduction:** A model `Input(10) -> Dense(64) -> Dense(32) -> Dense(3)` should report `2883` params. It currently reports `2179` (missing the `10*64+64 = 704` first-layer weights).

The existing test in `test_model_architecture.py` masks the bug because it uses list shapes (`shape: [10]`), which the real IR schema never produces.

## Expected behavior

The estimate should include the first layer's weights for scalar shapes: `2883` in the example above.

## Proposed fix

Add a scalar-int branch to the input-layer handler:

```python
elif isinstance(shape, int):
    prev_shape = shape
```

And add a regression test that uses the real IR-format scalar shape.

ashishsoni-ai:
## Summary

Fixes #402 — `GET /api/v1/model/architecture/{id}?include_stats=true` reported an incorrect parameter count for models whose input shape is stored as a scalar integer (the actual IR schema format).

### Backend

- **`deep_learning.py`** — `_estimate_param_count` now handles a scalar `int` input shape (the format the IR schema actually produces) in addition to the list format, so the first Dense layer's weights are counted correctly.

### Tests

- **`test_model_architecture.py`** — added a regression test using the real IR-format scalar shape (`Input(10) -> Dense(64) -> Dense(32) -> Dense(3)`) asserting `2883` params.

## Verification

- `test_model_architecture.py` → 8 passed
- `ruff check` clean

## Related

- Closes #402

### DIFF
--- a/tensormap-backend/app/services/deep_learning.py
+++ b/tensormap-backend/app/services/deep_learning.py
@@ -700,6 +700,8 @@ def _estimate_param_count(graph_ir: dict) -> int:
             if isinstance(shape, list) and len(shape) > 0:
                 # For input, store the last dimension
                 prev_shape = shape[-1] if shape[-1] is not None else 128  # default guess
+            elif isinstance(shape, int):
+                prev_shape = shape
 
         elif layer_type == "dense":
             units = node_params.get("units", 0)
--- a/tensormap-backend/tests/test_model_architecture.py
+++ b/tensormap-backend/tests/test_model_architecture.py
@@ -172,3 +172,29 @@ def test_estimate_param_count_embedding(db_session: Session):
     # Dense: 64 * 1 + 1 = 65
     # Total: 1,329,473
     assert data["param_count"] == 1329473
+
+
+def test_estimate_param_count_scalar_shape(db_session: Session):
+    """The IR schema stores input shape as a scalar int (e.g. 10), not a list.
+    The first Dense layer's weights must still be counted."""
+    graph_ir = {
+        "nodes": [
+            {"node_params": {"layer_type": "input", "shape": 10}},
+            {"node_params": {"layer_type": "dense", "units": 64}},
+            {"node_params": {"layer_type": "dense", "units": 32}},
+            {"node_params": {"layer_type": "dense", "units": 3}},
+        ]
+    }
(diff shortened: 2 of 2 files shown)


ivantha:
### Automated review summary

**Findings**
- None flagged.

**Conventions**
- _Title/commit subject ('fix: _estimate_param_count now handles scalar input shape...') is a bit long and descriptive rather than imperative (e.g. 'fix: handle scalar input shape in _estimate_param_count' would be tighter and avoid truncation in git log), but otherwise hygiene is clean._

This is a small, correctly-scoped bug fix: `_estimate_param_count` (tensormap-backend/app/services/deep_learning.py) previously only recognized a list-shaped `input` layer `shape`, so whenever the IR stored shape as a scalar int — the format the schema actually produces — `prev_shape` stayed `None` and the first Dense layer's weights were silently dropped from the parameter-count estimate. The added `elif isinstance(shape, int): prev_shape = shape` branch fixes this, and the new regression test (`Input(10) -> Dense(64) -> Dense(32) -> Dense(3)`, expecting 2883 params) exercises exactly the previously-broken path and matches manual calculation. The diff is minimal, behavior-preserving for the existing list-shape path, and well-verified; no correctness, security, or test-coverage issues found. Safe to merge as-is.

PR #405, opened 2026-09-02
fix: deterministic ordering for training-jobs endpoint and de-flake its test

### LINKED ISSUE #404: Flaky test: test_training_jobs_list_endpoint fails due to non-deterministic ordering of identical timestamps
## Summary

`test_training_jobs_list_endpoint` in `test_cp2_integration.py` is flaky: it creates three training jobs with identical `started_at` values and asserts they are returned most-recent-first. With no secondary sort key, a DB returns rows with identical `started_at` in non-deterministic order, so the test fails intermittently.

## Root cause

- The test uses `started_at=datetime.now(UTC)` for all three jobs — they land within the same microsecond.
- The endpoint `GET /model/training-jobs` orders by `started_at.desc()` only.
- The test asserts `jobs_data[0]["job_id"] == jobs[-1].id`, assuming insertion order, which is not guaranteed.

## Expected behavior

The endpoint should always return a deterministic, documented order (most recently started first), and the test should deterministically verify that behavior.

## Proposed fix

- `app/routers/training.py`: add `TrainingJob.id.desc()` as a secondary sort key so identical-`started_at` rows have a stable order.
- `tests/test_cp2_integration.py`: give the three jobs distinct, ascending `started_at` timestamps and assert the full expected order as a list of job IDs.

ashishsoni-ai:
## Summary

Fixes #404 — `test_training_jobs_list_endpoint` was flaky because three jobs with identical `started_at` timestamps had no deterministic ordering. The endpoint now uses a secondary sort key so the order is always stable, and the test uses distinct timestamps so the assertion is meaningful.

## Changes

- **`app/routers/training.py`** — add `TrainingJob.id.desc()` as a secondary sort key alongside `started_at.desc()` so rows with identical `started_at` values have a deterministic ordering.
- **`tests/test_cp2_integration.py`** — use distinct, ascending `started_at` timestamps (`now + timedelta(seconds=i)`) and assert the full expected order as a list of job IDs.

## Verification

- `test_cp2_integration.py` (single test) → 1 passed
- `test_cp2_integration.py` + `test_training_jobs.py` → 16 passed, 3 skipped
- `ruff check` clean

## Related

- Closes #404


ivantha:
### Automated review summary

**Findings**
- **[minor/test]** `tensormap-backend/tests/test_cp2_integration.py:350` The test now uses strictly increasing started_at values (base_time + timedelta(seconds=i)) for all three jobs, so no two jobs ever tie on started_at. This means the new TrainingJob.id.desc() secondary sort key added in training.py is never actually exercised by this test — the assertion would pass identically even if that tie-breaker were removed, since started_at.desc() alone already produces a unique, deterministic order here. The regression-test value for the specific fix (tie-breaking on equal timestamps) is effectively zero.

**Conventions**

The fix itself is correct and minimal: adding TrainingJob.id.desc() as a secondary ORDER BY key resolves genuinely non-deterministic ordering whenever two jobs share a started_at value. The test update (distinct ascending timestamps, asserting the full ordered ID list rather than just the first element) is a clear improvement over the old single-element assertion. The one gap is that the new timestamps never collide, so the test no longer exercises the tie-breaking path it's meant to guard — worth adding a case with identical started_at values to actually cover the id.desc() fallback. Title, description, and the single commit are all clear, conventional, and properly linked to the issue being closed.

PR #409, opened 2026-09-13
fix: reject values <= -1 in Log Transform preprocessing

### LINKED ISSUE #408: [Bug] Log Transform accepts -1 values, silently corrupts dataset with -inf
## Summary

The **Log Transform** preprocessing step only rejects values **below -1** (the check is `s < -1`), but `log(1 + x)` is undefined at **x = -1** as well: `log1p(-1)` evaluates to `-inf`.

As a result, a CSV column containing a value of exactly `-1` passes validation, is silently replaced with `-inf`, and the corrupted CSV **overwrites the user's original file** while the API responds with HTTP 200 (success). Infinite values then poison downstream training (NaN losses, broken metric plots).

## Steps to reproduce

1. Upload a CSV whose column contains a value of `-1`, e.g.

```csv
price
1.0
-1.0
3.0
```

2. Apply the "Log Transform" transformation to that column (`POST /data/process/preprocess/{file_id}`).
3. Observe: the request returns HTTP 200 (success) and the stored column now contains `-inf`.

```python
import numpy as np
import pandas as pd
from app.services.data_process import _log_transform

df = pd.DataFrame({"price": [1.0, -1.0, 3.0]})
out = _log_transform(df, "price", None)
print(out)
#       price
# 0  0.693147
# 1      -inf
# 2  1.386294
```

Note the `RuntimeWarning: divide by zero encountered in log1p` — and the request still succeeds.

## Expected behavior

The transformation should be rejected with a clear 422 error (like it already does for values below -1) when any value is **<= -1** (i.e. `log(1+x)` is undefined or non-finite), and the original file must remain untouched.

## Impact

- Silent data corruption (dataset overwritten with `-inf`).
- Downstream model training/evaluation can produce NaN losses and garbage metrics with no traceable cause.
- The bug is in `tensormap-backend/app/services/data_process.py` (`_log_transform`).

## Suggested fix

Change the guard from `s < -1` to `s <= -1` and include the offending column name/count in the error message.

ashishsoni-ai:
## What

Fixes #408: the **Log Transform** preprocessing step only rejected values strictly **below -1**, but `log(1+x)` is non-finite at `x == -1` (`log(0)` evaluates to `-inf`).

A CSV column containing a value of exactly `-1` therefore passed validation, produced `-inf`, and silently overwrote the user's uploaded file (HTTP 200) with corrupted data that poisons downstream training (NaN losses, garbage metrics).

## Changes

- `tensormap-backend/app/services/data_process.py` — `_log_transform` now guards on `s <= -1` instead of `s < -1`, and the error message names the offending column and count. Values `> -1` (e.g. `-0.999`) remain valid.
- `tensormap-backend/tests/test_new_transforms.py` — regression tests:
  - `test_rejects_value_of_exactly_negative_one` — a column containing `-1` returns 422 and the dataset is not mutated.
  - `test_accepts_values_just_above_negative_one` — values arbitrarily close to `-1` still transform to finite results.

## Verification

- `pytest tests/test_new_transforms.py tests/test_data_process.py tests/test_data_process_backup.py tests/test_data_process_correlation.py tests/test_data_process_params.py tests/test_data_process_service.py tests/test_column_stats.py` → **91 passed**
- `ruff check app/services/data_process.py tests/test_new_transforms.py` → **All checks passed!**

## Before / after

| Input (`price`) | Before | After |
|---|---|---|
| `[1.0, -1.0, 3.0]` | HTTP 200, column becomes `[0.693147, -inf, 1.386294]` | HTTP 422 with a clear message, file untouched |


ivantha:
### Automated review summary

**Findings**
- None flagged.

**Conventions**

This PR fixes a real correctness bug: `log1p(x)` is non-finite at `x == -1` (evaluates to `-inf`), and the previous `s < -1` check let such rows through, so a column containing exactly `-1` returned HTTP 200 while silently writing `-inf` into the user's CSV. Changing the guard to `s <= -1` closes that gap, and the two added tests correctly cover both the newly-rejected boundary value and a value just above it that must still transform. Validation happens before any file write in `preprocess_data`, so the PR's claim that the file is left untouched on rejection is accurate, and no other transform function in the file has the same boundary bug. This is a clean, minimal, well-tested fix ready to merge.

PR #412, opened 2026-09-18
fix: export SavedModel via Model.export() for Keras 3 compatibility

ashishsoni-ai:
## Summary

Fixes #411.

`export_savedmodel()` used the Keras 2 API `tf.keras.models.save_model(model, path, save_format="tf")`,
which Keras 3 rejects:

```
ValueError: The `save_format` argument is deprecated in Keras 3. Please remove this argument
and pass a file path with either `.keras` or `.h5` extension.
```

Since `tensormap-backend/pyproject.toml` pins `tensorflow>=2.16,<2.17` (which ships Keras 3), every
`GET /api/v1/model/export/{job_id}?format=savedmodel` request returned HTTP 500.

## Change

`app/services/model_export.py`
- replace `tf.keras.models.save_model(model, str(savedmodel_dir), save_format="tf")` with
  `model.export(str(savedmodel_dir))`, the supported Keras 3 way to write a SavedModel directory.
- zip packaging and the "already generated" cache behaviour are unchanged.

## Tests

`tests/test_model_export.py`
- `test_savedmodel_export_uses_model_export` (renamed from `..._creates_zip`): asserts `Model.export()`
  is called with the savedmodel directory and that `save_model()` is not called, so the Keras 2 call
  cannot silently come back.
- `test_savedmodel_export_produces_real_savedmodel_zip` (new, `slow`): real `save` → `export` → zip
  round-trip; asserts `saved_model.pb` is inside the archive and that a second call returns the cached
  archive.

Both new tests fail on the previous implementation with the exact `ValueError` above.

## Verification

```
cd tensormap-backend
python -m pytest tests/test_model_export.py -q
# 18 passed, 2 skipped

python -m ruff check app/services/model_export.py tests/test_model_export.py
# All checks passed!
python -m ruff format --check app/services/model_export.py tests/test_model_export.py
# 2 files already formatted
```

Also verified manually with `tensorflow==2.16.2` / `keras==3.13.2`: `model.export(dir)` produces
`saved_model.pb`, `variables/`, `assets/`, `fingerprint.pb`.

## Out of scope (separate issue #410)

- TFLite export is also broken on this stack, with a different root cause: TF 2.16's
  `TFLiteConverter.from_keras_model` calls `keras_deps.get_call_context_function()`, which is only
  registered by the Keras 2 `tf_keras` package.
- ONNX export is unavailable because `onnx 1.19` requires `ml_dtypes.float4e2m1fn` while TF 2.16 pins
  `ml_dtypes 0.3.2`.

PR #413, opened 2026-09-18
fix: convert TFLite from the SavedModel export (Keras 3 compatibility)

### LINKED ISSUE #410: Bug: TFLite and ONNX model export are broken on the pinned TensorFlow 2.16 / Keras 3 stack
## Summary

Two of the three model-download formats are dead on the backend stack this repo pins, so
"Export model" only ever works for SavedModel — and that one is broken too (#411).
Because the failures happen in native/TensorFlow code or at import time, the existing tests do not
notice (ONNX tests *skip*, TFLite tests mock the converter), so CI stays green.

1. `GET /api/v1/model/export/{job_id}?format=tflite` → HTTP 500 with
   `TypeError: 'NoneType' object is not callable`
2. `GET /api/v1/model/export/{job_id}?format=onnx` → ONNX is never available because
   `import tf2onnx` raises `AttributeError: module 'ml_dtypes' has no attribute 'float4e2m1fn'`

The API's own error message even tells users to fall back:
`"suggestion": "Use SavedModel or TFLite format instead"` — but TFLite is broken and SavedModel is
broken too.

## Environment

From `tensormap-backend/pyproject.toml` (`tensorflow>=2.16,<2.17`, `tf2onnx>=1.16`, `keras` unpinned),
resolved today to:

| Package      | Version |
|--------------|---------|
| tensorflow   | 2.16.2  |
| keras        | 3.13.2  |
| tf2onnx      | 1.17.0  |
| onnx         | 1.19.0  |
| ml_dtypes    | 0.3.2 (pinned transitively by TF 2.16) |

## 1. TFLite export fails (HTTP 500)

`app/services/model_export.py::export_tflite()` does:

```python
model = tf.keras.models.load_model(model_path)
converter = tf.lite.TFLiteConverter.from_keras_model(model)
tflite_model = converter.convert()
```

Minimal reproduction:

```python
import tensorflow as tf

model = tf.keras.Sequential([tf.keras.layers.Input((8,)), tf.keras.layers.Dense(4), tf.keras.layers.Dense(2)])
converter = tf.lite.TFLiteConverter.from_keras_model(model)
converter.convert()
```

```
  File ".../tensorflow/lite/python/tflite_keras_util.py", line 188, in _wrapped_model
    with keras_deps.get_call_context_function()().enter(
TypeError: 'NoneType' object is not callable
```

**Root cause:** TF 2.16's Keras-model → TFLite path calls
`tensorflow.python.util.keras_deps.get_call_context_function()`, which is only registered by the
legacy Keras 2 package (`tf_keras`, not installed here). This is a TensorFlow 2.16 + Keras 3
incompatibility (fixed in TF >= 2.17), not something TensorMap can fix by itself.

**Verified workarounds** on this exact stack (both produce a working `.tflite` buffer):

```python
# (a) convert from the SavedModel export
converter = tf.lite.TFLiteConverter.from_saved_model("<job>/savedmodel")

# (b) convert from concrete functions (no extra artifacts needed)
fn = tf.function(lambda x: model(x)).get_concrete_function(
    tf.TensorSpec(model.input_shape, tf.float32, name="input")
)
converter = tf.lite.TFLiteConverter.from_concrete_functions([fn], model)
```

Suggested fix: use (a) or (b) in `export_tflite()`, or bump TensorFlow to >= 2.17.

## 2. ONNX export is unavailable (and silently skipped)

```python
import tf2onnx
```

```
  File ".../onnx/_mapping.py", line 104, in <module>
    np.dtype(ml_dtypes.float4_e2m1fn),
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'ml_dtypes' has no attribute 'float4e2m1fn'
```

**Root cause:** TensorFlow 2.16 pins `ml_dtypes==0.3.2`, but `onnx 1.19` needs
`ml_dtypes.float4_e2m1fn` (added in ml_dtypes 0.4). Since `tf2onnx` is unpinned (`>=1.16`), a fresh
install resolves into a broken combination.

**Why it is invisible:** `tests/test_model_export.py::_check_tf2onnx_available()` returns `False`, so
the two ONNX tests `SKIP` and CI passes while the feature cannot work for any user.

Suggested fix: pin a compatible `tf2onnx`/`onnx`/`ml_dtypes` combination (or bump TensorFlow), and
make the ONNX unavailability show up as an explicit, visible signal instead of a silent skip.

## Impact

`GET /model/export/{job_id}/formats` reports per-format availability, and the frontend Export panel
offers all three downloads. On the pinned stack only SavedModel (after #411) can
actually be produced; TFLite and ONNX always fail.

ashishsoni-ai:
## Summary

Fixes #410 (TFLite half; see "ONNX" below).

`export_tflite()` used `tf.lite.TFLiteConverter.from_keras_model(model)`, which cannot convert Keras 3
models on the pinned stack (`tensorflow 2.16.2` + `keras 3.13.2`):

```
  File ".../tensorflow/lite/python/tflite_keras_util.py", line 188, in _wrapped_model
    with keras_deps.get_call_context_function()().enter(
TypeError: 'NoneType' object is not callable
```

TF 2.16 only registers that hook from the legacy Keras 2 package (`tf_keras`, not installed), so
every `GET /api/v1/model/export/{job_id}?format=tflite` returned HTTP 500. The API's own ONNX error
even recommends TFLite as the fallback ("Use SavedModel or TFLite format instead"), so users had no
working download at all.

## Change

`app/services/model_export.py`

1. `export_tflite()` now exports the model to the job's `savedmodel/` directory with
   `Model.export()` and converts with `TFLiteConverter.from_saved_model()` — the supported path.
   Verified it also covers **multi-input** models, where the concrete-function recipes do not.
2. `export_onnx()` no longer catches only `ImportError`. On this stack `import tf2onnx` raises
   `AttributeError` (onnx 1.19 needs `ml_dtypes.float4_e2m1fn`, while TF 2.16 pins `ml_dtypes 0.3.2`),
   which escaped the old handler and surfaced as a raw 500 with a cryptic message. It now raises
   `ONNXUnsupportedError`, which `app/routers/export.py` already maps to the documented
   400 `onnx_unsupported` payload — so the frontend gets a clear reason instead of a 500.

## Tests

`tests/test_model_export.py`

- `test_tflite_export_uses_savedmodel_path` (renamed from `test_tflite_export_creates_file`):
  asserts `from_keras_model()` is never called, the model is exported to the job's `savedmodel/`
  directory, and the converter is built from it.
- `test_tflite_export_produces_working_tflite` (new, `slow`): real `save` → `export` → `convert`,
  then loads the produced `.tflite` in a `tf.lite.Interpreter` and compares the output with eager
  Keras.
- `test_tflite_export_supports_multi_input_models` (new, `slow`): same for a two-input functional
  model (TFLite input ordering is not guaranteed, so inputs are matched by shape).
- `test_onnx_export_degrades_when_toolchain_unavailable` (new): simulates the `AttributeError`
  import failure and asserts `ONNXUnsupportedError` is raised with the real reason. Runs in CI even
  though the two existing ONNX tests are skipped there.

All four fail on the previous implementation.

## Verification

```
cd tensormap-backend
python -m pytest tests/test_model_export.py -q
# 20 passed, 2 skipped   (2 pre-existing ONNX skips)

python -m ruff check app/services/model_export.py tests/test_model_export.py
# All checks passed!
python -m ruff format --check app/services/model_export.py tests/test_model_export.py
# 2 files already formatted

python -m pytest tests/test_phase4_integration.py tests/test_cp2_integration.py tests/test_model_deletion_cascade.py -q
# 12 passed, 3 skipped
```

## Notes on fully enabling ONNX

Making ONNX *work* (rather than degrade gracefully) needs a dependency decision, because
`onnx<1.18` resolves `protobuf` into a version that clashes with the `protobuf>=4.25,<5` pin that
TensorFlow 2.16 requires. Suggested follow-ups (in a separate PR): bump TensorFlow to >= 2.17, or
coordinate an `onnx`/`ml_dtypes`/`protobuf` pin set. Related PR: #412 fixes the SavedModel export on
the same stack.


ivantha:
### Automated review summary

**Findings**
- **[major/bug]** `tensormap-backend/app/services/model_export.py:113` export_tflite() now calls model.export(str(savedmodel_dir)) against the exact same path, export_dir / "savedmodel", that export_savedmodel() (line 75) populates via tf.keras.models.save_model(model, str(savedmodel_dir), save_format="tf"). The two functions each have their own file-existence caching (tflite_path.exists() / zip_path.exists()) but neither checks whether the sibling export already created this shared directory. Since Keras 3's Model.export() writes a TF SavedModel artifact and is documented to fail (or behave unpredictably) when the target directory already exists and is non-empty, requesting a SavedModel export for a job and then a TFLite export for the same job (or vice versa) can throw instead of succeeding, even though both are legitimate, independently-cacheable downloads for the same trained model. _(unverified)_
- **[minor/quality]** `tensormap-backend/app/services/model_export.py:223` The tf2onnx import guard now does `except Exception as e` instead of `except ImportError`, which is the right fix for the AttributeError case described in the comment, but as written it will also swallow unrelated import-time failures (e.g. a genuine bug in tf2onnx's own module init, or a MemoryError/KeyboardInterrupt during import) and re-report them as an ordinary "toolchain unavailable" 400 response rather than surfacing them as real errors.

**Conventions**

This PR fixes a genuine bug: on the pinned TensorFlow 2.16 / Keras 3 stack, TFLiteConverter.from_keras_model() always raised TypeError, so every TFLite download returned HTTP 500; exporting to a SavedModel directory first and converting from that is the correct workaround, and it also hardens export_onnx()'s import guard to catch non-ImportError failures from a broken onnx/ml_dtypes toolchain. However, the new export_tflite() code writes its intermediate SavedModel to the exact same export_dir/savedmodel path that the pre-existing export_savedmodel() already uses, with no cross-function collision handling — requesting both export formats for one job can now fail instead of both succeeding. Test coverage is otherwise strong, including two slow tests that round-trip a produced .tflite file through a real TFLite interpreter, but there's no test exercising SavedModel export followed by TFLite export (or vice versa) for the same job, which is the scenario that would surface the directory collision. Commit hygiene and the PR description are clean and well justified.

